PR #6, opened 2017-10-05
kvdbproxy test + vpp-agent cn-infra update

lukasmacko:


lukasmacko:
Beware: after update to the latest cn-infra go1.9 is mandatory

PR #197, opened 2017-11-07
WIP: SNK 130,132: Basic Contiv Policy Generation, ADD/DEL events

brecode:
- Add/Del Policy
- Add/Update Pod

- Basic Contiv Policy Generation
  - No IP Blocks, No Namespace Selectors

Example of policy: 

apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: test-network-policy
  namespace: default
spec:
  egress:
    -
      ports:
        -
          port: 5978
          protocol: TCP
  ingress:
    -
      from:
        -
          podSelector:
            matchLabels:
              role: frontend
      ports:
        -
          port: 8080
          protocol: TCP
  podSelector:
    matchLabels:
      app: web
      role: db

Try adding Pods w/ Labels app:web, role:db and Pods w/ Labels role:frontend


milanlenco:
Merged, just couple of points to fix/discuss:
- use nsmodel.ID as namespace type in LookupPodsByNSLabelSelector()
- why to declare PolicyDefault if PolicyIngress is equivalent
- render even empty set of policies (to remove what is configured)
- do not return errors (just log) when waiting for more data from K8s
- can AddPod() receive all Pod data and not be followed by UpdatePod?
- UpdatePolicy will need to include pods that no longer match the policy (get from oldPolicy)

Some notes for the processor work to be done:
- for empty namespace selector (= all pods from all namespaces) set empty arrays Pods and IPBlocks in ContivPolicy (more efficient rules will be implemend)
- Filip will add GetHostSubnet() into Contiv to get the host's IP subnet; processor should only run Process() for pods within this subnet

PR #215, opened 2017-11-08
VCL namespace networking

lukasmacko:
Current limitation:
 - ping from VPP to pod doesn't work because applied stn rule/loopback


rastislavszabo:
The code needs to be updated to reflect new VPP binary api.

PR #242, opened 2017-11-09
SNK 190: Preparation for VPPTCP Renderer + Configurator UTs. 

milanlenco:


plugins/policy/configurator/configurator_test.go
```diff
@@ -0,0 +1,1207 @@
+package configurator
```
fgschwan: This file is one big monolith. Tests could be written more compact by extracting data structures away and making helper methods with good names. 
Basically this is just refactoring suggestion for which we don't have time right now, so ok for now.

PR #248, opened 2017-11-10
User cli-listen file to enable vppctl

hagbard5235:
Instructing folks to set cli-listen to 0.0.0.0:5002 is a big security
hole.  Change to using cli-listen to a socket file, so that folks can
use vppctl instead (which is secure).

Signed-off-by: Ed Warnicke <eaw@cisco.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,7 @@ to contain the proper PCI address:
 ```
 unix {
     nodaemon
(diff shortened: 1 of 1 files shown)


hagbard5235:
Need to correct other telnet related instructions.

PR #329, opened 2017-11-24
Issue 317: Use docker exec instead kubectl in test

vrpolakatcisco:
Signed-off-by: Vratko Polak <vrpolak@cisco.com>


vrpolakatcisco:
This does not work properly for 2node suites.

PR #349, opened 2017-11-28
Issue 317: Use docker exec instead kubectl in test

vrpolakatcisco:
Some changes to tweo node suite.

Signed-off-by: Vratko Polak <vrpolak@cisco.com>


vrpolakatcisco:
Commit message should be better.

PR #363, opened 2017-11-29
Tags10

vrpolakatcisco:


plugins/policy/plugin_impl_policy.go
```diff
@@ -157,13 +157,13 @@ func (p *Plugin) Init() error {
 	p.processor.Init()
 	p.configurator.Init(false) // Do not render in parallel while we do lot of debugging.
 	p.aclRenderer.Init()
-	if p.Contiv.IsTCPstackDisabled() {
+	if !p.Contiv.IsTCPstackDisabled() {
```
rastislavszabo: why the change in this file?

vrpolakatcisco: That comes from https://github.com/contiv/vpp/pull/362
I think when master is more recent that dev, it is ok to merge that changes to dev. (In the future I will be more careful so that git does not append such edits without my consent.)

rastislavszabo: right. yep, merging master -> dev at any time is allowed and welcome :)

PR #454, opened 2018-01-05
 Prototyping ksr restart resync - exploring building blocks DO NOT MERGE

jmedved:
Exploring how to implement ksr resync when ksr restarts; we will need to do something like:
- check that resync with k8s is finished
- read etcd state
- List k8s cache & mark sweep what's in etcs

We must avoid churning staeady state, i.e. if a value in etcd is not changing, we do not write. We need to update changed values, add new values and delete values no longer present in etcd. And we need to hold off any add/delete/update events until we're done.


jmedved:
Rasto, Lukas, Milan - can you guys please review, and if it's ok, merge? I only refactored the service reflector so far, the other ones stiff needs to be done. 

I have another question - how do we detect Etcd failure, and how do we recover from that?

plugins/ksr/ksr_reflector.go
```diff
@@ -0,0 +1,290 @@
...
+
+	kr.dsSynced = true
+}
+
+func (kr *Reflector) syncDataStore(dsItems DsItems, oc K8sToProtoConverter) {
+	for {
+		if kr.k8sController.HasSynced() {
+			break
+		}
+		time.Sleep(1 * time.Second)
```
rastislavszabo: I would put some smaller sleep time here (100ms ?)

lukasmacko:
ad etcd failure detection, 
currently it is possible to detect etcd connection failure using health/statuscheck plugin. It allows to retrieve current state of registered plugins and check if the etcd connectivity is ok. In order to recover properly I would suggest to add watcher API to statuscheck, that would notify subscribers on state change. Re-established etcd connection after failure would trigger resync.

jmedved:
Lukas, do you have an example of how to use the health/statuscheck plugin? Please point me to the code. 

In the meantime, can you guys please merge this patch, i would like to continue the work on a new patch:
- Add tests
- Migrate all other Reflectors onto the new structure (I started with Endpoints already, the Reflecto code is cleaned up quite a bit)

lukasmacko:
with the following patch applied to cn-infra
[statuscheckpatch.txt](https://github.com/contiv/vpp/files/1614524/statuscheck.txt)
you can read status of etcd using datasync example

PR #506, opened 2018-01-22
KSR data resync for failed etcd / etcd connection

jmedved:


plugins/ksr/plugin_impl_ksr.go
```diff
@@ -201,13 +208,75 @@ func (plugin *Plugin) AfterInit() error {
...
+	for {
+		select {
+		case <-closeCh:
+			plugin.Log.Info("Closing")
+			return
+		case <-time.After(1 * time.Second):
+			sts := plugin.StatusMonitor.GetAllPluginStatus()
+			for k, v := range sts {
+				if k != "etcdv3" {
+					return
```
lukasmacko: I think would be better to use either `continue` instead of `return` or map access `sts["etcdv3"]`. Otherwise, if AllPluginStatus contains status of a plugin different from etcdv3, monitoring is stopped.

PR #550, opened 2018-01-29
Removed = from type declaration to make linters happy

dseevr:
Signed-off-by: Bill Robinson <dseevr@users.noreply.github.com>


dseevr:
nvm, I see that these are used as alias types... the linter was incorrectly detecting it as an error

PR #555, opened 2018-01-31
Add vpp-vswitch docker build

tiewei:
Make vpp-vswitch image build using 2-stage build, in the builder
stage it requires "vpp-base-builder" image to compile due to the
vpp-dev.deb dependency, when compile completed, copy contiv-vpp
and configs onto the runner container.


rastislavszabo:
Hi,

the image misses ldpreload binaries:
https://github.com/contiv/vpp/blob/master/docker/ubuntu-based/prod/extract.sh#L38-L40

for reference, how it is being used:
https://github.com/contiv/vpp/blob/master/k8s/contiv-vpp.yaml#L206

tiewei:
@rastislavszabo  is `libvcl_ldpreload.so.0.0.0` the only file needed ? https://github.com/contiv/docker-vpp/blob/master/Dockerfile#L27 actually copies the whole directory over.
Also i have a question, why the `ldpreload.so` is put onto `/vpp-lib64/` (hostpath `/tmp/ldpreload/vpp-lib64`), is it actually being used ? both those two dir is not places that linux would find libs

rastislavszabo:
Hi,

yes, `libvcl_ldpreload.so.0.0.0` is the only file needed. 

We copy the file from the vswitch container to the host in the init container:
https://github.com/contiv/vpp/blob/master/k8s/contiv-vpp.yaml#L206

and then mount it from the host to the PODs by CRI, in case that TCP stack is enabled and the POD is LD_PRELOADed:

https://github.com/contiv/vpp/blob/master/pkg/manager/manager.go#L214

tiewei:
@rastislavszabo  https://github.com/contiv/docker-vpp/pull/9 should be able to address this issue once it merged. The binary will be put onto right place and the env var will be set. I don't think there will need change in the PR though

docker/vpp-vswitch/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# compile vpp-agent needs vpp libs
+FROM contiv/vpp-base-builder:b59bd659 as builder
```
dseevr: Will we be manually updating this SHA1?  Could this become an `ARG`?

docker/vpp-vswitch/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENV GOPATH /root/go
+ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+
+RUN apt-get update \
+    && apt-get install -y curl git make \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl -L "https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz" | tar -C /usr/local -xzf - \
+    && go get -u github.com/golang/lint/golint
+
+ADD . $GOPATH/src/github.com/contiv/vpp
```
dseevr: `COPY` is preferred over `ADD` unless you need the special stuff `ADD` can do:  https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy

docker/vpp-vswitch/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# lint and only build vpp-agent
+RUN cd $GOPATH/src/github.com/contiv/vpp \
+    && make lint agent
+
+FROM contiv/vpp-base:b59bd659
```
dseevr: Same here... can this be the same `ARG` as the builder image?

docker/vpp-vswitch/Dockerfile
tiewei: I thought ADD will do a compression+transport , looks like i misread the doc, will fix

docker/vpp-vswitch/Dockerfile
tiewei: Technically we could, but that needs to create a build hook in order to let docker cloud know which is the right image version and pass-in as build-args, however, that hook dir impacts all images built in this repo, so I think for now we just keep it in the dockerfile, it's not supposed to be changed frequently anyway

dseevr: Maybe just add a note that if you change the sha1 fragment here, you also need to change it down below too

docker/vpp-vswitch/govpp.conf
```diff
@@ -0,0 +1,3 @@
+health-check-probe-interval: 1000000000
+health-check-reply-timeout: 500000000
+health-check-threshold: 1
```
unclejack: This needs a line at the end of the file. ;)

docker/vpp-vswitch/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY . $GOPATH/src/github.com/contiv/vpp
+
+# lint and only build vpp-agent
+RUN cd $GOPATH/src/github.com/contiv/vpp \
+    && make lint agent
+
+FROM contiv/vpp-base:b59bd659
+
+RUN apt-get update \
+    && apt-get install -y ethtool supervisor
```
unclejack: Perhaps you could remove the apt lists here to make the image a bit smaller?

docker/vpp-vswitch/govpp.conf
tiewei: dang, i thought i fixed it. ..

tiewei: done

docker/vpp-vswitch/Dockerfile
tiewei: done

PR #562, opened 2018-01-31
added `metalinter` make target which runs gometalinter in a container

dseevr:
Signed-off-by: Bill Robinson <dseevr@users.noreply.github.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,6 +104,12 @@ define format_only
     @echo "# done"
 endef
 
+# run all code quality tools in parallel
+define metalinter_only
+    @docker build -t vpp_metalinter -f ./docker/development/Dockerfile.metalinter .
+    @docker run --rm vpp_metalinter
+endef
+
 # build contiv agent
 define build_contiv_agent_only
     @echo "# building contiv-agent"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -104,6 +104,12 @@ define format_only
     @echo "# done"
 endef
 
+# run all code quality tools in parallel
+define metalinter_only
+    @docker build -t vpp_metalinter -f ./docker/development/Dockerfile.metalinter .
+    @docker run --rm vpp_metalinter
```
tiewei: can we instead build the vpp metalinter and just run it ?

dseevr: I'm not sure what you mean... these two lines are building and running it 🤔

tiewei: i mean we don't build it on the fly, it doesn't look we will need to rebuild it everytime, so why not we just push it to dockerhub and pull it every time ?

dseevr: It has to `COPY` in the code every time so we do need to build it every time.  All of the major parts are cached, though.  It's just the `COPY` + `WORKDIR` + `ENTRYPOINT` that gets built.  This is how all the "checks" Dockerfiles work across all our repos (at least the ones I've been involved with 😄 ).

The other option is to use bindmounts, but those have the disadvantages we already discussed so I think we should continue to avoid them in general.

tiewei: fair enough

Issue #558, opened 2018-01-31
Static route from host to service subnet


rastislavszabo:

jmedved:
Issue: Kubernetes service can not be accessed through its Cluster IP from a remote host. The service can only be accessed from a local host. The service can still be accessed from Pods running on remote hosts. To be fully compliant with K8s, we need to be able to access any service from any host.

PR #564, opened 2018-02-01
Add ServiceCIDR to IPAM config + add route from host in order to acce…

lukasmacko:
#558


tiewei:
Please brief describe what is #558 and why this fixes #558 in commit msg

jmedved:
This patch fixes issue #558 (https://github.com/contiv/vpp/issues/558).

The solution is to add a default CIDR subnet from which Cluster IPs are allocated and to install a route on every host pointing to this subnet.

PR #565, opened 2018-02-01
3 fixes for the ACLs.

milanlenco:


tiewei:
Please brief why the fixes are needed, and what it's fixing.
I can tell the `cache_impl.go` one is because `LookupPod` already returned `false, nil`, the 2nd update is trying to do a (deep?) copy before change ? could you have more info in your commit msg ?
ref:
https://chris.beams.io/posts/git-commit/
https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

jmedved:
This pull request fixes a problem where ACLs for an ingress policy applied to a server pod were not correctly deleted when a client pod accessing the server was terminated.

The problem was caused by not performing a deep copy of ACLs passed to the ACL renderer, causing them to be modified while they were being rendered.

PR #567, opened 2018-02-02
Update docs for ServiceCIDR option

lukasmacko:


k8s/README.md
```diff
@@ -49,7 +49,8 @@ To use the development image for testing with specific version of VPP, see
       on 1 k8s node (VPPHost network = VPPHost subnet for one k8s node);
     - `NodeInterconnectCIDR`: subnet used for main interfaces of all nodes;
     - `VxlanCIDR`: subnet used for VXLAN addressing providing node-interconnect overlay.
-    - `ServiceCIDR`: subnet used for allocation of Cluster IPs for services
+    - `ServiceCIDR`: subnet used for allocation of Cluster IPs for services. Default value
+    is default kubernetes service range `10.96.0.0/12`
```
jmedved: ... is the default kubernetes service range ... (add 'the')

k8s/README.md
```diff
@@ -49,7 +49,8 @@ To use the development image for testing with specific version of VPP, see
       on 1 k8s node (VPPHost network = VPPHost subnet for one k8s node);
     - `NodeInterconnectCIDR`: subnet used for main interfaces of all nodes;
     - `VxlanCIDR`: subnet used for VXLAN addressing providing node-interconnect overlay.
-    - `ServiceCIDR`: subnet used for allocation of Cluster IPs for services
+    - `ServiceCIDR`: subnet used for allocation of Cluster IPs for services. Default value
+    is the default kubernetes service range `10.96.0.0/12`.
```
tiewei: i think this default comes from kubeadm default settings,  i think we should uncomment `ServiceCIDR: "10.96.0.0/12"` in yaml file and add comments that you will need to update it to match kubeadm `--service-cidr` value if it's set

PR #568, opened 2018-02-02
Policy processor refactoring + properly remove ACLs for removed pods

milanlenco:
The pull request contains refactor of processor.go from the policy plugin, focused on avoiding copy&paste code. The file size went down by 300 lines just by putting repeated code into functions.
Additionally, multiple bugs affecting policies were fixed:
- service and policy plugin do both watch for changes in pods (as reflected by KSR) but they cannot do it from the same kvdbsync plugin, at least not until this is supported by cn-infra (something for the backlog of the ligato team). For now they will have their own copies of the plugin.
- last cleanup from Nikos has brough some small bugs with it (cache_impl.go and match_label.go)
- removal  of a pod that had some ACL attached was not handled correctly at all. A small change was required in the processor - now we uninstall policies as soon as the pod looses its IP address. Furthermore the configurator and the renderer has to remember pod IPs and interface names, respectivelly, since the uninstallation of policies happen after the pod was removed and these information are no longer known by the Contiv plugin or the policy cache.

### DIFF
--- a/plugins/policy/cache/cache_impl.go
+++ b/plugins/policy/cache/cache_impl.go
@@ -143,8 +143,27 @@ func (pc *PolicyCache) LookupPodsByNSLabelSelector(policyNamespace string,
 // LookupPodsByLabelSelector evaluates label selector (expression and/or match
 // labels) and returns IDs of matching pods.
 func (pc *PolicyCache) LookupPodsByLabelSelector(
-	podLabelSelector *policymodel.Policy_LabelSelector) (pods []podmodel.ID) {
-	return nil
+	namespaceLabelSelector *policymodel.Policy_LabelSelector) (pods []podmodel.ID) {
+	// An empty namespace selector matches all namespaces.
+	if len(namespaceLabelSelector.MatchExpression) == 0 && len(namespaceLabelSelector.MatchLabel) == 0 {
+		allPods := pc.configuredPods.ListAll()
+		kubeSystemPods := pc.configuredPods.LookupPodsByNamespace("kube-system")
+		pods := utils.Difference(allPods, kubeSystemPods)
+
+		pc.Log.WithField("LookupPodsByNSLabelSelector", namespaceLabelSelector).
+			Infof("Empty namespace selector returning pods: %+v", pods)
+		return utils.UnstringPodID(pods)
+	}
+	pc.Log.WithField("LookupPodsByNSLabelSelector", namespaceLabelSelector).
+		Debugf("Namespace Label Selector is: %+v", namespaceLabelSelector)
+	// List of match labels and match expressions.
+	matchLabels := namespaceLabelSelector.MatchLabel
+
+	found, namespaceSelectorPods := pc.getPodsByLabelSelector(matchLabels)
+	if !found {
+		return []podmodel.ID{}
+	}
+	return utils.UnstringPodID(namespaceSelectorPods)
 }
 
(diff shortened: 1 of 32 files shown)


milanlenco:
Fixes dangling ACLs + 09-allow-traffic-only-to-a-port-number.md
Will re-check every test once again.

plugins/policy/cache/cache_impl.go
```diff
@@ -107,8 +107,13 @@ func (pc *PolicyCache) LookupPodsByNSLabelSelector(policyNamespace string,
...
 
-	return utils.UnstringPodID(utils.Intersect(mlPods, mePods))
-
+	if len(matchLabels) > 0 && len(matchExpressions) > 0 {
+		return utils.UnstringPodID(utils.Intersect(mlPods, mePods))
+	}
+	if len(matchLabels) > 0 {
+		return utils.UnstringPodID(mlPods)
+	}
+	return utils.UnstringPodID(mePods)
```
jmedved: A little nit:  the name of the function should be utils.UnstringPodIDs (with and 's' at the end to indicate plural)

plugins/policy/cache/cache_impl.go
```diff
@@ -240,8 +245,9 @@ func (pc *PolicyCache) LookupNamespace(namespace nsmodel.ID) (found bool, data *
 // LookupNamespacesByLabelSelector evaluates label selector (expression
 // and/or match labels) and returns IDs of matching namespaces.
 func (pc *PolicyCache) LookupNamespacesByLabelSelector(
-	nsLabelSelector string) []string {
-	return pc.configuredNamespaces.LookupNamespacesByLabelSelector(nsLabelSelector)
+	nsLabelSelector string) []nsmodel.ID {
+	namespaces := pc.configuredNamespaces.LookupNamespacesByLabelSelector(nsLabelSelector)
+	return utils.UnstringNamespaceID(namespaces)
 }
```
jmedved: A little nit:  the name of the function should be utils. UnstringNamespaceIDs (with and 's' at the end to indicate plural)

plugins/policy/configurator/configurator_impl.go
```diff
@@ -130,45 +137,61 @@ func (pct *PolicyConfiguratorTxn) Commit() error {
...
+			}
+			pct.podIPAddresses[pod] = podIPNet
+
+			// Sort policies to get the same outcome for the same set.
+			policies := unorderedPolicies.Copy()
+			sort.Sort(policies)
+
+			// Check if this set was already processed.
+			alreadyProcessed := false
+			for _, policySet := range processed {
```
tiewei: This is trying to remove duplicated policies, i think we could do a pre-processing of `unorderedPolicies` to make it in order and without duplication, but it's fine for now

plugins/policy/configurator/configurator_impl.go
```diff
@@ -545,6 +561,15 @@ func (cr ContivRules) Copy() ContivRules {
 	return crCopy
 }
 
+// Copy creates a deep copy of PodIPAddresses.
+func (pa PodIPAddresses) Copy() PodIPAddresses {
+	paCopy := make(PodIPAddresses, len(pa))
+	for pod, addr := range pa {
+		paCopy[pod] = addr
```
tiewei: I didn't quite follow why we needs to hold a shallow copy, you might want deep copy if you want no reference back to old object

PR #566, opened 2018-02-02
Add support for namespace label selectors in K8s policies

brecode:
This PR makes a few changes in the policy plugin, adding support for namespace label selectors, both for Ingress and Egress Rules. Also fixes bugs related to empty podSelectors and empty namespaceLabelSelectors. Fixes should be also part of #560 and #565

PR #570, opened 2018-02-03
Policy Cache & Policy Processor fixes for #566

brecode:
This PR addresses comment suggestions of #566.

 - Refactor code for match_expression and match_label go files to return only []string
    instead of (bool, []string)

 - Simplified getPodsByNSLabelSelector & getPodsByLabelSelector in match_label.go to be
   easily readable

 - Added missing comments for function description and more robust comments in code

 - Fixed two bugs in processor.go (missing "!" and worng function call)

### DIFF
--- a/plugins/policy/cache/match_expression.go
+++ b/plugins/policy/cache/match_expression.go
@@ -29,26 +29,26 @@ const (
 )
 
 // getMatchExpressionPods returns all the pods that match a collection of expressions (expressions are ANDed)
-func (pc *PolicyCache) getMatchExpressionPods(namespace string, expressions []*policymodel.Policy_LabelSelector_LabelExpression) (bool, []string) {
+func (pc *PolicyCache) getMatchExpressionPods(namespace string, expressions []*policymodel.Policy_LabelSelector_LabelExpression) []string {
 	var inPodSet, notInPodSet, existsPodSet, notExistPodSet []string
 	for _, expression := range expressions {
 		switch expression.Operator {
 		case in:
 			labels := utils.ConstructLabels(expression.Key, expression.Value)
-			isMatch, podSet := pc.getPodsByNSLabelSelector(namespace, labels)
-			if !isMatch {
-				return false, nil
+			podSet := pc.getPodsByNSLabelSelector(namespace, labels)
+			if len(podSet) == 0 {
+				return []string{}
 			}
 
 			inPodSet = append(inPodSet, podSet...)
 
 		case notIn:
 			labels := utils.ConstructLabels(expression.Key, expression.Value)
-			isMatch, podSet := pc.getPodsByNSLabelSelector(namespace, labels)
-			if !isMatch {
+			podSet := pc.getPodsByNSLabelSelector(namespace, labels)
+			if len(podSet) == 0 {
 				podNamespaceAll := pc.configuredPods.LookupPodsByNamespace(namespace)
-				if podNamespaceAll == nil {
-					return false, nil
+				if len(podNamespaceAll) == 0 {
+					return []string{}
 				}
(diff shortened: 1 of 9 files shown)


plugins/policy/cache/match_expression.go
```diff
@@ -97,17 +97,14 @@ func (pc *PolicyCache) getMatchExpressionPods(namespace string, expressions []*p
...
+	if len(existsMatcher) == 0 {
+		return []string{}
 	}
 
 	pods := utils.Intersect(inMatcher, existsMatcher)
-	if pods == nil {
-		return false, nil
-	}
-	return true, pods
+	return pods
```
tiewei: same here, I would say update `utils.Intersect` to make it smarter on either of them is empty case, like 
```
func Intersect(a []string, b []string) []string {
   if len(a) == 0 || len(b) == 0 {
          return []string{}
   }
    ...
}
```
or further improve it to take a slice variables
```
// Intersect at least need two lists
func Intersect(a []string, b []string, s ...[]string) []string {
	if len(a) == 0 || len(b) == 0 {
		return []string{}
	}
	set := make([]string, 0)
	hash := make(map[string]bool)
	for _, el := range a {
		hash[el] = true
	}
	for _, el := range b {
		if _, found := hash[el]; found {
			set = append(set, el)
		}
	}
	if len(s) == 0 {
		return set
	}
	return Intersect(set, s[0], s[1:]...)
}
```
In this way, *a lot* places can be just
```
return utils.Intersect(inPodSet, notInPodSet, existsPodSet, notExistPodSet)
```

plugins/policy/cache/match_label.go
```diff
@@ -21,66 +21,44 @@ import (
...
 	}
 
-	return true, newPodSet
+	return current
 }
 
-// GetPodsByNSLabelSelector returns the pods that match a collection of Label Selectors
-func (pc *PolicyCache) getPodsByLabelSelector(labels []*policymodel.Policy_Label) (bool, []string) {
+// getPodsByLabelSelector returns the pods that match a collection of Label Selectors
+func (pc *PolicyCache) getPodsByLabelSelector(labels []*policymodel.Policy_Label) []string {
```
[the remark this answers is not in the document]
brecode: done

plugins/policy/cache/match_label.go
```diff
@@ -21,66 +21,44 @@ import (
 	"github.com/contiv/vpp/plugins/policy/utils"
 )
 
-// GetPodsByNSLabelSelector returns the pods that match a collection of Label Selectors in the same namespace
-func (pc *PolicyCache) getPodsByNSLabelSelector(namespace string, labels []*policymodel.Policy_Label) (bool, []string) {
-	newPodSet := []string{}
-
+// getPodsByNSLabelSelector returns the pods that match a collection of Label Selectors in the same namespace
+func (pc *PolicyCache) getPodsByNSLabelSelector(namespace string, labels []*policymodel.Policy_Label) []string {
```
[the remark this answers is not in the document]
brecode: done

plugins/policy/utils/utils.go
```diff
@@ -41,21 +41,26 @@ func RemoveDuplicates(el []string) []string {
 	return result
 }
 
-// Intersect returns the common elements of two slices
-func Intersect(a []string, b []string) []string {
+// Intersect returns the common elements of or more slices
```
tiewei: a typo: `two or more slices`

plugins/policy/utils/utils.go
```diff
@@ -41,21 +41,26 @@ func RemoveDuplicates(el []string) []string {
...
+// Intersect needs at least two lists
```
tiewei: if corrected above line, this line is not needed

PR #573, opened 2018-02-06
Change node-local IPAM to not reuse IPs immediately

lukasmacko:
This patch changes the IP allocation algorithm in node-local IPAM.  The goal of this change is to avoid applying policies to wrong pods. Currently, if a Pod that satisfies a given policy exits and a new Pod that does not satisfy the policy is started a short time after, the new Pod will get the previous Pod's IP address and the policy will be erroneously applied to the new Pod.

The solution is to keep allocating previously unused IP addresses from the local IPAM range. Previously used pod IP addresses are only reused only if the all available range is exhausted.

### DIFF
--- a/plugins/contiv/ipam/ipam.go
+++ b/plugins/contiv/ipam/ipam.go
@@ -53,6 +53,8 @@ type IPAM struct {
 	nodeInterconnectCIDR net.IPNet // IPv4 subnet used for for inter-node connections
 	vxlanCIDR            net.IPNet // IPv4 subnet used for for inter-node VXLAN
 	serviceCIDR          net.IPNet // IPv4 subnet used to allocate ClusterIPs for a service
+
+	lastAssigned int // counter denoting last assigned IP address
 }
 
 type uintIP = uint32
@@ -73,8 +75,9 @@ type Config struct {
 func New(logger logging.Logger, nodeID uint8, config *Config) (*IPAM, error) {
 	// create basic IPAM
 	ipam := &IPAM{
-		logger: logger,
-		nodeID: nodeID,
+		logger:       logger,
+		nodeID:       nodeID,
+		lastAssigned: 1,
 	}
 
 	// computing IPAM struct variables from IPAM config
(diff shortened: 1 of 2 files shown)


plugins/contiv/ipam/ipam.go
```diff
@@ -243,29 +246,50 @@ func (i *IPAM) NextPodIP(podID string) (net.IP, error) {
...
 	return nil, fmt.Errorf("No IP address is free for assignment. All IP addresses for pod network %v are already assigned", i.podNetworkIPPrefix)
 }
 
+// tryToAllocatePodIP checks whether the IP at the given index is available.
+func (i *IPAM) tryToAllocatePodIP(index int, networkPrefix uint32, podID string) (assignedIP net.IP, success bool) {
+	if index == podGatewaySeqID {
+		return nil, false // gateway IP address can't be assigned as pod
+	}
+	ip := networkPrefix + uint32(index)
+	if _, found := i.assignedPodIPs[ip]; found {
```
tiewei: i don't see `assignedPodIPs` is initiated from etcd, how you make sure it's not being used after restarting ?

plugins/contiv/ipam/ipam.go
```diff
@@ -243,29 +246,50 @@ func (i *IPAM) NextPodIP(podID string) (net.IP, error) {
...
+		return nil, false // ignore already assigned IP addresses
+	}
+	i.assignedPodIPs[ip] = podID
```
tiewei: you might want to add lock for the `assignedPodIPs` read/write or even for the allocate ip function, it should be safe to do so b/c you are using RLock

plugins/contiv/ipam/ipam.go
```diff
@@ -243,29 +246,50 @@ func (i *IPAM) NextPodIP(podID string) (net.IP, error) {
...
-		i.assignedPodIPs[networkPrefix+uint32(j)] = podID
-		//TODO set etcd for new assigned value
+	}
 
-		ipForAssign := uint32ToIpv4(networkPrefix + uint32(j))
-		i.logger.Infof("Assigned new pod IP %s", ipForAssign)
-		i.logAssignedPodIPPool()
-		return ipForAssign, nil
+	// iterate from the range start until lastAssigned
+	for j := 1; j < last; j++ { // zero ending IP is reserved for network => skip seqID=0
```
tiewei: you might want to wrap the try assign as a method or just put it as part of `tryToAllocatePodIP`, so that in case we want to change the algorithm again.

plugins/contiv/ipam/ipam.go
```diff
@@ -73,8 +75,9 @@ type Config struct {
 func New(logger logging.Logger, nodeID uint8, config *Config) (*IPAM, error) {
 	// create basic IPAM
 	ipam := &IPAM{
-		logger: logger,
-		nodeID: nodeID,
+		logger:       logger,
+		nodeID:       nodeID,
+		lastAssigned: 1,
```
tiewei: you might want to init it with the gw index

PR #575, opened 2018-02-06
Add support for configurable default route (default GW) 

rastislavszabo:
This PR removes the hardcoded default route from VPP to the host stack. Instead, 
it allows specifying a custom default gateway as a node-config option `Gateway` 
in the contiv-vpp.yaml deployment file.

In order to remove the default route to host on VPP, specific routes need to be added:
VPP needs to be able to reach host's local IP addresses (addresses on interfaces not bound to VPP),
since one of them may be the main IP address of the node, as k8s sees it. Since this IP is used
for NAT translations of services with host networking, PODs behind VPP need to be able to reach it.

Note: "github.com/vishvananda/netlink" should be added as an explicit dependency once https://github.com/contiv/vpp/pull/574 is merged. Not required for successful build, since it already exists in the vendor folder due to transient dependencies.


k8s/README.md
```diff
@@ -60,7 +60,8 @@ To use the development image for testing with specific version of VPP, see
       - `IP`: IP address to be attached to the main interface;
     - `OtherVPPInterfaces` (other configured interfaces only get IP address assigned in VPP)
       - `InterfaceName`: name of the interface;
-      - `IP`: IP address to be attached to the interface.
+      - `IP`: IP address to be attached to the interface;
+    - `Gateway`: IP address of the default gateway for external traffic, if it needs to be configured.
```
chrisplo: shouldn't gateway be a child of MainVppInterface and OtherVPPInterfaces or are all vpp interfaces requirdd to be on the same network to expand bandwidth (not familiar with multiple VPP interface configurations)

plugins/contiv/host.go
```diff
@@ -79,18 +80,34 @@ func (s *remoteCNIserver) routeServicesFromHost() *linux_l3.LinuxStaticRoutes_Ro
...
+		NextHopAddr:       gwIP,
+		OutgoingInterface: outIfName,
 	}
 	return route
 }
 
+func (s *remoteCNIserver) routesToHost() []*vpp_l3.StaticRoutes_Route {
+	// list all IPs assigned to host interfaces
+	ips, err := s.getHostLinkIPs()
+	if err != nil {
```
chrisplo: logging of the error might be helpful here

plugins/contiv/host.go
```diff
@@ -292,3 +309,29 @@ func (s *remoteCNIserver) ipPrefixToAddress(ip string) string {
...
+				return nil, err
+			}
+			// return all IPs
+			for _, addr := range addrList {
+				res = append(res, addr.IP.String())
+			}
+		}
+	}
+	return res, nil
+}
```
chrisplo: is there any way to detect if host IPs change (CCP primary interface is DHCP for example, or maybe a down interface comes up) to make route adjustments?

An error might be good too if a host IP overlaps one of the IPAM networks?

plugins/contiv/plugin_impl_contiv.go
```diff
@@ -90,9 +90,11 @@ type Config struct {
...
 // OneNodeConfig represents configuration for one node. It contains only settings specific to given node.
 type OneNodeConfig struct {
-	NodeName           string
-	MainVppInterface   InterfaceWithIP
-	OtherVPPInterfaces []InterfaceWithIP // other configured interfaces get only ip address assigned in vpp
+	NodeName           string            // name of the node, should match withs the hostname
+	MainVppInterface   InterfaceWithIP   // main VPP interface used for the inter-node connectivity
+	OtherVPPInterfaces []InterfaceWithIP // other interfaces on VPP, not necessarily used for inter-node connectivity
+	Gateway            string            // IP address of the default gateway
+	StealHostInterface string            // interface in the host Linux (e.g. eth0) to be stolen and connected to VPP
```
chrisplo: shouldn't StealHostInterface be part of another PR?

k8s/README.md
rastislavszabo: I think Gateway is not an interface-local setting, but a VPP-global setting. I can point to an IP behind any of the VPP interfaces (managed by OtherVPPInterfaces). But still, you should have only one default gateway (one default route) in the setup.

plugins/contiv/plugin_impl_contiv.go
rastislavszabo: right, is an future option. Removed.

plugins/contiv/host.go
rastislavszabo: log added

plugins/contiv/host.go
rastislavszabo: This is pretty-much a temporary solution (ETA to be removed is 1 week). To fix another issue, we need to ask k8s what k8s thinks is the main IP address of our node. This will be done via k8s state reflector (therefore needs some more time to implement). I added a TODO explaining this. I think that addressing these corner cases is not worth of time investment at this point.

plugins/contiv/host.go
```diff
@@ -79,18 +80,35 @@ func (s *remoteCNIserver) routeServicesFromHost() *linux_l3.LinuxStaticRoutes_Ro
...
+	if err != nil {
+		s.Logger.Errorf("Error by listing host IPs: %v", err)
+		return nil
+	}
+
+	// generate a /32 static route from VPP for each of the host's IPs
+	routes := make([]*vpp_l3.StaticRoutes_Route, 0)
+	for _, ip := range ips {
+		routes = append(routes, &vpp_l3.StaticRoutes_Route{
+			DstIpAddr:   fmt.Sprintf("%s/32", ip),
```
tiewei: this assumes ipv4, could be a future improvment

plugins/contiv/host.go
tiewei: yeah getting from node internal ip is more reasonable

k8s/README.md
chrisplo: right, thx

jmedved:
I'll try to finish the required KSR changes today or not later than tomorrow.

PR #577, opened 2018-02-07
Create deep copy of BD config

lukasmacko:
The PR fixes the issue of accessing pods deployed on a different node if the vxlans are used. Vxlan was not added to bridge domain.

The configuration of bridge domain where vxlan tunels are terminated is stored in the remote cni server. When a new node joins the cluster the config is updated and passed to local client. Unfortunately, (due to performance) local client does not create a copy of the configuration. Thus, before this fix we were effectively overwriting prev value of the change event generated by local client, i.e prevValue and new value were identical. Despite, the overwriting everything used to work as expected because BD was recreated upon each modification. Latest vpp-agent modifies BD only if neccessery (i.e. if there is a change which was the case because of overwrite).

Solution implemeted in the PR is to pass a deep copy to local client in order to avoid overwriting the prev value.

### DIFF
--- a/plugins/contiv/node_events.go
+++ b/plugins/contiv/node_events.go
@@ -22,7 +22,9 @@ import (
 	"net"
 
 	"github.com/contiv/vpp/plugins/contiv/model/node"
+	"github.com/golang/protobuf/proto"
 	"github.com/ligato/cn-infra/datasync"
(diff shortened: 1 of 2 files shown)


rastislavszabo:
Merging, since this is a critical bug fix.

PR #579, opened 2018-02-08
docker/vpp-cni: fix static binary build

unclejack:
This commit fixes the build of the contiv-cni binary. It wasn't static before.


docker/vpp-cni/Dockerfile
```diff
@@ -14,7 +14,7 @@ WORKDIR /go/src/github.com/contiv/vpp/cmd/contiv-cni
 
 # we collect & store all files in one folder to make the resulting
 # image smaller when we copy them all in one single operation
-RUN CGO_ENABLED=0 \
+RUN export CGO_ENABLED=0 \
```
vhosakot: Does `RUN export` persist across images? I've seen issues with it. Why not use `ENV` here instead?

vhosakot:
👍

PR #580, opened 2018-02-08
Vagrantfile: make a few small changes & fixes

unclejack:
This PR makes the following changes:
- changes some messages to say exactly what's being done
- adds support for vagrant-cachier
- enables kvm paravirt
- bumps Go to 1.9.3

The recently released Go 1.9.4 isn't used because it might break some packages which rely on CGO.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -64,17 +64,16 @@ export https_proxy='#{https_proxy}'
 EOF
 
 source /etc/profile.d/envvar.sh
-echo "Updating Ubuntu..."
+echo "Updating apt lists..."
 apt-get update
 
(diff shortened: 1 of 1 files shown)


vagrant/Vagrantfile
```diff
@@ -234,13 +233,19 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
     config.vm.provider 'virtualbox' do |v|
         v.linked_clone = true if Vagrant::VERSION >= "1.8"
+        v.customize ['modifyvm', :id, '--paravirtprovider', 'kvm']
```
tiewei: This only works on hosts has kvm supported

PR #581, opened 2018-02-09
Fixes for DeletePod ACL cleanup and Namespace Label updates

brecode:
This PR addresses the following issues:

- ACLs were not cleanup up properly after a delete pod event. If the ip address of the pod is not removed on the update state,
then pod is sent to configurator and policies are not properly removed if the interface exists (or remain the same). Old DelPod function on
the processor.go would return nil. Updated DelPod function will send the deleted pod and all the affected pods to the configurator.

- Added update namespace labels support. On namespace edit, if there is a change in the labels, UpdateNamespace function gets called
and finds all the pods that are affected from the change (pods that are affected should only be the ones in Egress/Ingress namespace selectors)

### DIFF
--- a/plugins/policy/processor/processor.go
+++ b/plugins/policy/processor/processor.go
@@ -156,8 +156,33 @@ func (pp *PolicyProcessor) AddPod(podID podmodel.ID, pod *podmodel.Pod) error {
 
 // DelPod processes the event of a removed pod (no action needed).
 func (pp *PolicyProcessor) DelPod(podID podmodel.ID, pod *podmodel.Pod) error {
-	pp.Log.WithField("podID", podID).Info("Pod was removed")
-	/* Already un-configured when the pod has lost its IP address */
+	pp.Log.WithFields(logging.Fields{
+		"podID":   podID,
+		"del-pod": pod,
+	}).Info("Pod was removed")
+	if pod.Namespace == "kube-system" {
+		pp.Log.WithField("pod", pod).Info("Pod belongs to kube-system namespace, ignoring")
+		return nil
+	}
+
+	pods := []podmodel.ID{}
+	podPolicies := pp.getPoliciesAssignedToPod(pod)
+	for _, policy := range podPolicies {
+		pods = append(pods, pp.getPodsAssignedToPolicy(policy)...)
+	}
+	strPods := utils.RemoveDuplicates(utils.StringPodID(pods))
+	pods = utils.UnstringPodID(strPods)
+
+	// Re-configure only pods that belong to the current node.
+	hostPods := pp.filterHostPods(pods)
+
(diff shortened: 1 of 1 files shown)


plugins/policy/processor/processor.go
```diff
@@ -462,3 +512,71 @@ func (pp *PolicyProcessor) getPoliciesAssignedToPod(pod *podmodel.Pod) (policies
...
+		} else {
+			for _, ingressRules := range dataPolicy.IngressRule {
+				for _, ingressRule := range ingressRules.From {
+					matchLabels := []*policymodel.Policy_Label{}
+
+					if ingressRule.Namespaces != nil {
+						matchLabels = ingressRule.Namespaces.MatchLabel
+						for _, label := range matchLabels {
+							if labelsMap[label.Key] != label.Value {
+								break
```
tiewei: looks like you added `policies[dataPolicyID] = dataPolicy` anyway, why need this for loop at all ?

plugins/policy/processor/processor.go
```diff
@@ -462,3 +512,71 @@ func (pp *PolicyProcessor) getPoliciesAssignedToPod(pod *podmodel.Pod) (policies
...
+					}
+					policies[dataPolicyID] = dataPolicy
+				}
+			}
+		}
+
+		if len(dataPolicy.EgressRule) == 0 {
+			continue
+		} else {
+			for _, egressRules := range dataPolicy.EgressRule {
```
tiewei: this 3-level for loop can be moved out for a method to avoid repeating yourself

plugins/policy/processor/processor.go
```diff
@@ -156,8 +156,33 @@ func (pp *PolicyProcessor) AddPod(podID podmodel.ID, pod *podmodel.Pod) error {
...
 // DelPod processes the event of a removed pod (no action needed).
 func (pp *PolicyProcessor) DelPod(podID podmodel.ID, pod *podmodel.Pod) error {
-	pp.Log.WithField("podID", podID).Info("Pod was removed")
-	/* Already un-configured when the pod has lost its IP address */
+	pp.Log.WithFields(logging.Fields{
+		"podID":   podID,
+		"del-pod": pod,
+	}).Info("Pod was removed")
+	if pod.Namespace == "kube-system" {
+		pp.Log.WithField("pod", pod).Info("Pod belongs to kube-system namespace, ignoring")
```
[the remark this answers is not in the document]
brecode: We ignore policy processing for for kube-system pods

plugins/policy/processor/processor.go
brecode: Yup, policies[dataPolicyID] = dataPolicy should be in the for loop, fixed

PR #583, opened 2018-02-09
Add node reflector to KSR

jmedved:
This pull request adds to the KSR the ability to distribute a subset of node data to Contiv-VPP vswitches. This functionality is implemented in a new component - the Node Reflector. 

The Node Reflector is built on top of the base KSR Reflector. The new components are as follows:
- plugins/ksr/model/node: A new folder that contains the protobuf definition and the generated code for node data that is stored in Etcd
- plugins/ksr/node_reflector.go: contains the initialization code, functions for Adding/Deleting/Updating node data and the converted of node data from K8s format into the format used in Etcd.

This patch also contains a minor cleanup of how reflectors are started.

Only Commits from Feb 07 and Feb 08 are a part of this patch.


plugins/ksr/model/node/node.proto
```diff
@@ -0,0 +1,108 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// Package node defines data model for Kubernetes Node.
+package node;
+
+// Pod is a collection of containers that can run on a host.
```
tiewei: "Node is a ..."

PR #584, opened 2018-02-09
DHCP support

lukasmacko:
This PR allows to configure DHCP on the main VPP interface. Renewal of DHCP lease is not supported.


k8s/README.md
```diff
@@ -54,6 +54,8 @@ To use the development image for testing with specific version of VPP, see
 
   * Node configuration (section `NodeConfig`; one entry for each node)
     - `NodeName`: name of a Kubernetes node;
+    - `UseDhcpOnMainInt`: acquire IP address for the main VPP interface using DHCP 
+        (beware: renew of IP address is not supported)
```
tiewei: is renew lease or changing ip ?

jmedved: The IP address could be changed when the lease is renewed. I guess that's what Lukas is referring to. If the lease is renewed, but the IP address stays the same, we should be fine.

plugins/service/plugin_impl_service.go
```diff
@@ -173,6 +174,16 @@ func (p *Plugin) watchEvents() {
...
+		nodeIP := p.Contiv.GetNodeIP()
+		if nodeIP != nil {
+			break
+		}
+		// TODO: remove sleep i.e: extend API of contiv plugin
+		time.Sleep(500 * time.Millisecond)
+	}
+
 	for {
 		select {
```
jmedved: Should you be getting the IP address through DHCP also in the Resync case?

k8s/README.md
lukasmacko: yes correct, the change of IP address is not supported. I'll re-phrase it

PR #594, opened 2018-02-15
Add routes to managementIP of other nodes

lukasmacko:
The aim of this PR is to fix the access to services running on the master where backend has host networking from a worker. For this type of services k8s management IP is used as the endpoint. 

Soulution: Each node exposes its management IP via etcd. When a node joins cluster other nodes will add route to the management IPs of the node.


plugins/contiv/host.go
```diff
@@ -261,6 +261,13 @@ func (s *remoteCNIserver) routeToOtherHostStack(hostID uint8, nextHopIP string)
 	return s.routeToOtherHostNetworks(hostNw, nextHopIP)
 }
 
+func (s *remoteCNIserver) routeToOtherManagementIP(managementIP string, nextHopIP string) *vpp_l3.StaticRoutes_Route {
+	return &vpp_l3.StaticRoutes_Route{
+		DstIpAddr:   managementIP + "/32",
+		NextHopAddr: nextHopIP,
+	}
```
jmedved: We will need to make the IP address generic IPv6/IPv4; we'll mrege for now, but IP addressing needs to be fixed overall.

PR #597, opened 2018-02-15
Secure VPP debug access

nbartos:
The current method of listening on a TCP socket allows non-root
users on the local host, and remote users to alter the network
configuration.  By switching this to a UNIX socket that only root
can access on the local host, we remove this vulnerability.

Note:  The documented netcat client does not support tab completion.
In the future, we should be able to make a simple CLI program that
will work with tab completion.

Drive by:  Cleaned up an inline bash script so it now reports error
status correctly, and made it more readable.

### DIFF
--- a/docker/vpp-vswitch/contiv-vswitch.conf
+++ b/docker/vpp-vswitch/contiv-vswitch.conf
@@ -1,6 +1,6 @@
 unix {
     nodaemon
-    cli-listen 0.0.0.0:5002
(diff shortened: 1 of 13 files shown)


nbartos:
Note:  the 'o pipefail' was removed in this changeset as compared to the previous one.

brecode:
Looks, good. Tested with Vagrant dev, works fine. Merging

PR #599, opened 2018-02-17
Added VMWare Fusion support for vagrant. Cleaned up Vagrantfiles

brecode:
* Vagrantfile templates are stored under Vagrantfiles folder.
  Option to choose between a production/development cluster and vmware/virtualbox provider.
* Cleaned up Vagrantfiles:
  - Removed hardcoded IP addresses, hostnames.
  - Deleted unnecessary package installations on production environment.
* Modified vagrant-up script to support both providers.

### DIFF
--- a/vagrant/README.md
+++ b/vagrant/README.md
@@ -53,13 +53,13 @@ leave empty) or the *development environment* (use the -d flag) . Instructions
 on how to build the development contiv/vpp-vswitch image can be found in the
 next paragraph.
 
-_For the testing environment run:_
+For the testing environment run:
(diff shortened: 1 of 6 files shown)


unclejack:
@brecode: This approach doesn't look good to me. We already have 4 Vagrantfiles to keep in sync. We should only have one Vagrantfile with some if conditions. This is going to be pretty hard to maintain. It's already been pretty difficult to deal with.

unclejack:
I had some comments left on this which weren't addressed in any way.

PR #600, opened 2018-02-20
Add unit tests to the statscollector plugin

jmedved:
Also, the code was slightly refactored for better testability,
and a bug was fixed where stats entry was not deleted when a
pod was deleted.

### DIFF
--- a/plugins/statscollector/plugin_impl_statscollector.go
+++ b/plugins/statscollector/plugin_impl_statscollector.go
@@ -138,11 +138,11 @@ func (p *Plugin) processPodEvent(event containeridx.ChangeEvent) {
 	if !event.Del {
 		return
 	}
-	nsmap, exists := p.podIfs[event.Value.PodName]
+	nsmap, exists := p.podIfs[event.Value.PodNamespace]
 	if !exists {
 		return
 	}
-	ifs, exists := nsmap[event.Value.PodNamespace]
+	ifs, exists := nsmap[event.Value.PodName]
 	if !exists {
 		return
 	}
--- a/plugins/statscollector/plugin_impl_statscollector.go
+++ b/plugins/statscollector/plugin_impl_statscollector.go
@@ -39,6 +39,8 @@ const (
 	outErrorPacketsMetric = "outErrorPackets"
 )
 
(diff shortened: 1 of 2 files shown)


plugins/statscollector/plugin_impl_statscollector.go
```diff
@@ -126,45 +128,49 @@ func (p *Plugin) Init() error {
...
-						interfaceNameLabel: entry.data.Name,
-					})
-					if err != nil {
-						p.Log.Error(err)
-					}
+	var err error
+	if !event.Del {
+		return
+	}
+	nsmap, exists := p.podIfs[event.Value.PodName]
```
lukasmacko: first index should be namespace see https://github.com/contiv/vpp/blob/acc1665ab8af2c4e5d7a593f905b1ae6bee0052f/plugins/statscollector/plugin_impl_statscollector.go#L49

and the function where the values are actualy inserted 
https://github.com/contiv/vpp/blob/acc1665ab8af2c4e5d7a593f905b1ae6bee0052f/plugins/statscollector/plugin_impl_statscollector.go#L249-L255

plugins/statscollector/plugin_impl_statscollector.go
```diff
@@ -126,45 +128,49 @@ func (p *Plugin) Init() error {
...
+	if !exists {
+		return
+	}
+	ifs, exists := nsmap[event.Value.PodNamespace]
```
lukasmacko: second index should be podName

plugins/statscollector/plugin_statscollector_test.go
```diff
@@ -0,0 +1,403 @@
...
+// GetNsIndex returns the index of the VPP session namespace associated
+// with the given pod.
+func (mc *mockContiv) GetNsIndex(podNamespace string, podName string) (nsIndex uint32, exists bool) {
+	return 0, false
+}
+
+func (mc *mockContiv) GetPodByIf(ifname string) (podNamespace string, podName string, exists bool) {
+	pod, found := mc.pods[ifname]
+	if found {
+		return pod.name, pod.namespace, true
```
lukasmacko: return values are swapped it should be `return pod.namespace, pod.name, true`

PR #605, opened 2018-02-22
docker: build images using new Dockerfiles

unclejack:
This change makes the scripts build images basedon the new Dockerfiles.

The new images which are built using the new associated Dockerfile are:
- vpp-cni
- vpp-ksr
- vpp-cri

The other images are left as they are. Those need to be handled
separately.

Please keep in mind that I haven't removed the old Dockerfiles and their associated dev/prod directories. We need to check if the automated builds rely on those individual scripts. If they do rely on those individual scripts, I have to break up the build-new.sh script into multiple scripts which are stored in the directory of every image.

### DIFF
--- a/docker/build-all.sh
+++ b/docker/build-all.sh
@@ -49,6 +49,6 @@ done
 cd ubuntu-based
 ./build.sh
 
-# builds all Alpine Linux -based images
-cd ../alpine-based
-./build.sh
+# builds the new images (vpp-cni, vpp-ksr, vpp-cri)
+cd ..
+./build-new.sh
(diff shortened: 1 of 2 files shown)


rastislavszabo:
Hi,

what are the requirements for running the new builds? I cannot build the locally:

```
$ sudo docker build -t prod-contiv-cni --no-cache --force-rm -f docker/vpp-cni/Dockerfile .
Sending build context to Docker daemon 422.2 MB
Step 1/9 : FROM golang:1.9.3-alpine3.7 as builder
Error parsing reference: "golang:1.9.3-alpine3.7 as builder" is not a valid repository/tag: invalid reference format
```
Docker version 1.13.1, build 092cba3

unclejack:
@rastislavszabo: You need Docker 17.06-ce or newer. This is needed for multi-stage builds.

PR #609, opened 2018-02-23
[WIP, READY FOR REVIEW] Integration of policies with services and the Internet access

milanlenco:
**WORK IN PROGRESS: PLEASE DO NOT MERGE YET**

**TO-BE-DONE:**
- **end-to-end testing** (none done yet, just via UTs; proof of concept done manually via VPP CLI)
- **documentation** (algorithm explanation, diagrams)

This pull request primarily includes the **refactor of the policy rendering code**, which was necessary to **adapt to the limitations of the VPP/NAT plugin**. For policies we always need to evaluate rules against the _local_ IP addresses and not the NATed addresses of services or the node itself. This is, however, not the case for inbound ACLs, which makes them unusable. Previously we were using both directions, but now we combine ingress with egress and install all rules into outbound ACLs. Furthermore, to apply access control on the inter-node and pod-to-internet traffic, we need to reflect the ingress policies into a "global" ACL, installed on the node's output interfaces, also outbound side.
A detailed algorithm description + diagrams depicting the order of VPP nodes will be part of the documentation.

Similar restrictions are also present in the VPPTCP stack - each pod has only a single "local" table of rules assigned (evaluated in the ingress direction) and the stack additionally provides a single "global" table, evaluated in the ingress direction for traffic entering the node.
The equivalent limitations of VPPTCP stack and ACL+NAT (just different orientation of tables) have allowed us to **unify the cache and the rendering algorithm to a large degree between the two renderers**. That's the second contribution of the pull request.

The third contribution is in the service plugin: the plugin now also installs **SNAT configuration which allows Internet access from pods**. The SNAT is configured on the physical interface which acts as the default GW (host-VPP interconnect is not supported for Internet access). The implementation is DHCP-aware.
The only issue is that we cannot SNAT inter-node traffic, otherwise policies on the destination node will be evaluated against the NATed address of the source node and not the source pod. The solution is to split inter-node traffic from the pod-internet traffic. This is possible with VXLANs (inter-node traffic is encapsulated. whereas pod-internet is not), or by having an additional physical interface which acts as the default GW. With VXLANs disabled and only one physical interface available, SNAT gets therefore disabled (and needs to be performed by and external NAT device). This is a limitation for which we don't have a workaround at the moment.

Both policies and services (+SNAT) have resync fully implemented, i.e. restart scenarious are supported.


brecode:
LGTM

PR #610, opened 2018-02-23
docker/,virtualbox-prod: export & use built images

unclejack:
This change makes it possible to use built images in the production environment.

This makes it possible to test with images built locally without pulling.

### DIFF
--- a/docker/save.sh
+++ b/docker/save.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+# Copyright (c) 2017 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


docker/save.sh
```diff
@@ -0,0 +1,37 @@
...
+docker tag prod-contiv-vswitch:$tag contivvpp/vswitch:latest
+
+if [ -f ../vagrant/images.tar ]; then
+	rm ../vagrant/images.tar
+fi
+
+docker save $images -o ../vagrant/images.tar
+
+for img in $images; do
+	docker rmi $img:latest
```
brecode: docker rmi $img

Do you need the ":latest" suffix?

unclejack: @brecode: I've fixed it. PTAL

brecode:
LGTM

PR #612, opened 2018-02-26
Unconfigure pod fix the order of delete

lukasmacko:
If the order is incorrect there are errors and cni sends multiple request to undeploy the pod.
Correct order:
- routes, arp (config items that refers to auto_TAP)
- auto_TAP ineterface (config item that refers to TAP created by VPP)
- vpp tap


milanlenco:
Merging as I want to include this in the ongoing policy testing.

PR #617, opened 2018-02-28
Add a vppctl shim for vpp debug access

nbartos:


rastislavszabo:
Hi,
I don't like 2 things on this pull request:
- why is this script coming to the CNI docker image? It has nothing to do with CNI, in fact, it also refers to the vswitch docker  image. To correct place for this script is, therefore, contivvpp/vswitch.
- it copies the script named `vppctl` into `/usr/local/bin` on the host, which would conflict with the original `vppctl` binary in case it is installed on the host. I would rather name it differently, e.g. `cvppctl`, `vswitchctl` or so

nbartos:
I put it in the CNI image because that's where other similar file installations were being done, but you are correct it really should be in the vswitch image.  I do like using the standard vppctl name though.  I am actually wondering if I could statically compile vppctl so it didn't require the extra library, and just copy the binary itself to /usr/local/bin/.  That would probably be cleaner.  I will explore if that is possible.

PR #625, opened 2018-03-01
VPP-side POD interface configurable CIDR and invalid VPP-side interface address fix

brecode:
This PR provides a way to configure through the contiv-vpp yaml file the
VPP-side POD interface CIDR.

Also fixes a bug where the same interface would get invalid addresses (>>255)
With this fix, addresses on the vpp-side (tap or afpacket) interface will mirror the last octet
of the POD ip address while using the subnet configured in the contiv-vpp yaml file.

### DIFF
--- a/plugins/contiv/ipam/ipam.go
+++ b/plugins/contiv/ipam/ipam.go
@@ -181,8 +181,11 @@ func (i *IPAM) VPPHostNetwork() *net.IPNet {
 }
 
 // VPPIfIPPrefix returns VPP-side interface IP address prefix.
-func (i *IPAM) VPPIfIPPrefix() net.IP {
-	return i.podIfIPCIDR.IP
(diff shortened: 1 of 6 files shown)


plugins/contiv/ipam/ipam.go
```diff
@@ -175,6 +180,11 @@ func (i *IPAM) VPPHostNetwork() *net.IPNet {
 	return &vSwitchNetwork
 }
 
+// VPPIfIPPrefix returns VPP-side interface IP address prefix.
+func (i *IPAM) VPPIfIPPrefix() net.IP {
+	return i.podIfIPCIDR.IP
+}
+
 // OtherNodeVPPHostNetwork returns VPP-host network of another node identified by nodeID.
```
jmedved: Need to do the defensive copy here similar to what is in the functions above

PR #634, opened 2018-03-05
Fix stn-install.sh to work with the new Dockerfile

rastislavszabo:


unclejack:
What is broken? How do we check?

rastislavszabo:
executing stn-install.sh was returning an error, since it was relying on the location of the STN binary. In the new version of Docker, the binary is in `/stn`, whereas in the old one, it was `/root/contiv-stn`. Since the entrypoint is set properly in both versions, specifying the path is not neccessary and fixes the issue.

PR #635, opened 2018-03-05
Do not reconfigure vswitch connectivity if it is persisted

lukasmacko:
If a worker node is restarted the vswitch configuration is reconfigured by resync procedure. In this case it is desired to not reconfigure & persist the config again.

### DIFF
--- a/plugins/contiv/remote_cni_server.go
+++ b/plugins/contiv/remote_cni_server.go
@@ -178,6 +178,10 @@ type remoteCNIserver struct {
 
 // vswitchConfig holds base vSwitch VPP configuration.
 type vswitchConfig struct {
+	// configured if set to true denotes that vswitch configuration is applied by resync of default plugins
+	// the local client txns are not executed only local variables are filled in order to provide correct values by getters
+	configured bool
+
 	nics         []*vpp_intf.Interfaces_Interface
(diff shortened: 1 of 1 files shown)


plugins/contiv/remote_cni_server.go
```diff
@@ -675,7 +693,9 @@ func (s *remoteCNIserver) configureVswitchHostConnectivity(config *vswitchConfig
 		// finish TAP configuration
 		if s.useTAPInterfaces {
 			config.tapHost = s.interconnectTapHost()
-			err = s.vppTxnFactory().Put().LinuxInterface(config.tapHost).Send().ReceiveReply()
+			if !config.configured {
+				err = s.vppTxnFactory().Put().LinuxInterface(config.tapHost).Send().ReceiveReply()
+			}
 			if err != nil {
 				s.Logger.Error(err)
 				return err
```
jmedved: It might be better to do:
+			if !config.configured {
+				err := s.vppTxnFactory().Put().LinuxInterface(config.tapHost).Send().ReceiveReply()
+				if err != nil {
+ 	 				s.Logger.Error(err)
+	  				return err
+				}
+			}

lukasmacko: Since we return everytime we encounter an error I didn't want to introdouce too many nested levels. But ok I'll change that

PR #637, opened 2018-03-06
Add MTU Size support and Configuration edit through Helm charts

brecode:
This PR adds support to make MTU size configurable through Helm charts and contiv-vpp.yaml. 

Added MTU field on Af_packet and tap interface configuration to support different MTU sizes. 
Field was added to Pod-VPP and interconnectAfpacket/interconnectTapHost interfaces

### DIFF
--- a/k8s/contiv-vpp/templates/vpp.yaml
+++ b/k8s/contiv-vpp/templates/vpp.yaml
@@ -25,6 +25,7 @@ data:
     TAPInterfaceVersion: {{ .Values.contiv.tapInterfaceVersion }}
     {{- if .Values.contiv.stealTheNIC }}
     StealTheNIC: True
+    MTUSize: {{ .Values.contiv.mtuSize }}
     {{- end }}
     IPAMConfig:
       PodSubnetCIDR: {{ .Values.contiv.ipamConfig.podSubnetCIDR }}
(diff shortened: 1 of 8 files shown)


k8s/contiv-vpp/templates/vpp.yaml
```diff
@@ -25,6 +25,7 @@ data:
     TAPInterfaceVersion: {{ .Values.contiv.tapInterfaceVersion }}
     {{- if .Values.contiv.stealTheNIC }}
     StealTheNIC: True
+    MTUSize: {{ .Values.contiv.mtuSize }}
     {{- end }}
     IPAMConfig:
```
jmedved: MTUSize needs to move past Line 29

plugins/contiv/pod.go
```diff
@@ -241,6 +245,7 @@ func (s *remoteCNIserver) podTAP(request *cni.CNIRequest, podIPNet *net.IPNet) *
 	return &linux_intf.LinuxInterfaces_Interface{
 		Name:    "pod-" + s.tapTmpHostNameFromRequest(request),
 		Type:    linux_intf.LinuxInterfaces_AUTO_TAP,
+		Mtu:     s.config.MTUSize,
```
jmedved: A question - where is the MTU set for the interface to the Linux stack?

brecode: Yes, check host.go changes

k8s/contiv-vpp/templates/vpp.yaml
brecode: Yup, quick copy-paste mistake

rastislavszabo:
Nikos, please also cover the configuration in https://github.com/contiv/vpp/blob/master/cmd/contiv-init/vppcfg.go - to support the STN case.

PR #647, opened 2018-03-08
Persist configIndex

lukasmacko:
- modfied data structure that is stored in configIndex. Only names that are necessary to remove config items are persisted.
- containerID is used as identifier in configIndex and IPAM isntead of networkNamespace because remove request sent after restart does not contain network namespace


plugins/contiv/containeridx/persist_test.go
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License.
+
+package containeridx
+
+import (
+	"github.com/contiv/vpp/mock/broker"
+	"github.com/contiv/vpp/plugins/contiv/containeridx/model"
+	"github.com/ligato/cn-infra/core"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/onsi/gomega"
```
ondrej-fabry: Could use `. "github.com/onsi/gomega` for more readable code, without `gomega.` prefixes all over the place.

PR #660, opened 2018-03-12
Bump vpp

nbartos:


docker/ubuntu-based/prod/vswitch/Dockerfile
```diff
@@ -1,15 +1,15 @@
 FROM ubuntu:16.04
 
 RUN apt-get update \
- && apt-get install -y \
+    && apt-get install -y \
```
unclejack: @nbartos: Can you leave these as space-&&, please? This is how I've formatted some other Dockerfiles. This makes them easier to read when all the && line up under the RUN commands.

rastislavszabo:
The VPP version cannot be bumped without updating ligato/vpp-agent in the vendor folder, which must be binary compatible with the version of VPP we want to bup to.

PR #662, opened 2018-03-13
docker/ubuntu-based: move vpp to separate image

unclejack:
This change makes it possible to build VPP and tag it with the commit ID. This will enable us to reuse a built VPP Docker image. This image will get rebuilt whenever the VPP git commit is changed, when the build-vpp.sh script is modified or when the Dockerfile is modified.

This also drops some `--no-cache` arguments which don't make sense. We should rely on Docker's invalidation and address caching issues another way.

This change helps speed up the build process.


rastislavszabo:
Hi, could you please resolve the conflict, so that we can merge this?

unclejack:
I've rebased the PR. The build is fine even when building from the latest tag. We can update this further when needed.

@rastislavszabo: PTAL

PR #667, opened 2018-03-13
 Use github.com/vpp-dev/vpp/tree/stable-1801-contiv as the base code for VPP

rastislavszabo:
This MR migrates the code base for VPP to github.com/vpp-dev/vpp/tree/stable-1801-contiv
It also contains update of the ligato/vpp-agent, which is now compatible with that version of VPP.
This version of VPP contains new NAT fixes, that should resolve some issues with STN deployment with SNAT.


docker/ubuntu-based/dev/build-vpp.sh
```diff
@@ -3,9 +3,14 @@
 set -euo pipefail
 
 cd /opt/vpp-agent/dev/
-git clone https://gerrit.fd.io/r/vpp
+
+#git clone https://gerrit.fd.io/r/vpp
```
unclejack: Perhaps these commented lines can go away?

docker/build-all.sh
```diff
@@ -18,7 +18,7 @@ set -e
 
 # default values for build args and VPP commit ID
 export DOCKER_BUILD_ARGS=""
-export VPP_COMMIT_ID="3a5938223395bb6dc74769d109471f4210217b43"
+export VPP_COMMIT_ID=""
```
unclejack: Why is this not needed any more? How will this affect the build?

docker/ubuntu-based/dev/build-vpp.sh
rastislavszabo: later on, we may go back to this location, but yes, is is not needed

docker/build-all.sh
rastislavszabo: This means we will always build VPP with the newest code from github.com/vpp-dev/vpp/tree/stable-1801-contiv .
Since we control that repo, we don't need to hardcode the commit ID here. The build should not be affected in any way, since this commit ID should be used only by pulling the VPP code.

rastislavszabo: Ah, I see that with regard to your proposed MR https://github.com/contiv/vpp/pull/662/ we may actually need it to have it set?

unclejack: @rastislavszabo I can modify PR #662 to take this into account. It's not a big deal. This was more of a question to understand what's going on. I can add some code to fetch the commit ID and use that when building the dev-vswitch image.

PR #708, opened 2018-04-01
 Fix vppctl for K8s 1.10

jmedved:
In K8s 1.10 only Image IDs are displayed by 'docker ps' instead of Image names. The vppctl script is modified to try to get an Image ID for the contivvpp/vswitch image name and then use the Image ID
to try to get the running vswitch container.

### DIFF
--- a/docker/ubuntu-based/prod/vswitch/vppctl
+++ b/docker/ubuntu-based/prod/vswitch/vppctl
@@ -4,12 +4,11 @@
 
 set -euo pipefail
 
-get_vswitch_container_ID() {
+get_vswitch_container_id() {
     # There may be more than one vswitch container if an upgrade/redeploy is
     # happening, so go for the newest one.
-    CID=$(docker ps --format '{{.CreatedAt}} {{.Image}} {{.ID}}' | grep -E " $1@?[^ ]* [^ ]+$" | sort -nr | \
-        head -n 1 | sed -e 's|.* ||g') || [ -z "$CID" ]
-    echo "$CID"
+    docker ps --format '{{.CreatedAt}} {{.Image}} {{.ID}}' | grep -E " $1@?[^ ]* [^ ]+$" | sort -nr | \
+        head -n 1 | sed -e 's|.* ||g'
 }
 
 IMAGE="contivvpp/vswitch"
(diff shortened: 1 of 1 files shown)


docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -4,6 +4,14 @@
 
 set -euo pipefail
 
+get_vswitch_container_ID() {
+    # There may be more than one vswitch container if an upgrade/redeploy is
+    # happening, so go for the newest one.
+    CID=$(docker ps --format '{{.CreatedAt}} {{.Image}} {{.ID}}' | grep -E " $1@?[^ ]* [^ ]+$" | sort -nr | \
+        head -n 1 | sed -e 's|.* ||g') || [ -z "$CID" ]
```
nbartos: The:
|| [ -z "$CID" ] 
Shouldn't be here.  Additionally, there is no point setting a local variable then echoing it.  Just do:
docker ps --format '{{.CreatedAt}} {{.Image}} {{.ID}}' | grep -E " $1@?[^ ]* [^ ]+$" | sort -nr | \
    head -n 1 | sed -e 's|.* ||g'

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -4,6 +4,14 @@
 
 set -euo pipefail
 
+get_vswitch_container_ID() {
```
nbartos: Having ID caps at the end looks weird, it's not really consistent.  I'd prefer lower case.

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -13,12 +21,28 @@ then
     exit 1
 fi
 
-# There may be more than one vswitch container if an upgrade/redeploy is happening, so go for the newest one.
-if ! ID=$(docker ps --format '{{.CreatedAt}} {{.Image}} {{.ID}}' | grep -E " $IMAGE@?[^ ]* [^ ]+$" | sort -nr | \
-    head -n 1 | sed -e 's|.* ||g') || [ -z "$ID" ]
-then
-    echo "Error finding a running container derived from the $IMAGE image." >&2
-    exit 1
+ID=$(get_vswitch_container_ID "$IMAGE")
```
nbartos: If this command fails, it will kill the script, which is probably not what you want.

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -13,12 +21,28 @@ then
...
+
+# In K8s 1.10 only Image IDs are displayed by 'docker ps' instead of Image
+# names. If we can't get the vswitch container by its name, try using Image
+# ID, which we get from listing docker images and looking for the latest
+# vswitch image.
+if [ ! "$ID" ] ; then
```
nbartos: You probably want to wrap these together like:
if ! ID=$(get_vswitch_container_ID "$IMAGE") || [ -z "$ID" ]

I'd prefer you keep the format consistent with what is already in the file, e.g. put "then" on the next line without a semicolon.

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -13,12 +21,28 @@ then
...
+    # Just grab the latest (by date) image for now. Will have to be replaced
+    # later with grabbing an image with a specific tag.
+    IMG=$(docker images --format '{{.CreatedAt}};{{.ID}};{{.Repository}}' | grep "$IMAGE" | sort -nr | head -n 1) \
```
nbartos: Grepping for $IMAGE without delimiters could give you a match that you don't want.

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -13,12 +21,28 @@ then
...
+        || [ -z "$IMG" ]
+    if [ "$IMG" ] ; then
```
nbartos: You probably want to combine these as I the other example I gave above.

docker/ubuntu-based/prod/vswitch/vppctl
```diff
@@ -13,12 +21,28 @@ then
...
+        IFS=';' read -ra IMG_FIELDS <<< "$IMG"
+        ID=$(get_vswitch_container_ID "${IMG_FIELDS[1]}")
+        if [ "$ID" == "" ] ; then
```
nbartos: It's better to use the -n and -z tests.

docker/ubuntu-based/prod/vswitch/vppctl
jmedved: This is just repeating what the script was doing before. If the '|| [ -z "$CID" ]' is not there, and there is no container present, the script will fail. If i don't put '|| [ -z "$CID" ]' here, i need to put it in all places where this function is called (as you point out later). Is that what you prefer?

docker/ubuntu-based/prod/vswitch/vppctl
jmedved: See above - where should the test for unbounded variable be put - inside the function or everywhere where the function is used?

docker/ubuntu-based/prod/vswitch/vppctl
jmedved: what delimiters would you suggest? The output line format is very simple, just 3 fields, and none of them is remotely similar to the image name string.

PR #718, opened 2018-04-05
Update vpp-agent and increase timeout for bin API response to 3sec

milanlenco:


k8s/contiv-vpp.yaml
```diff
@@ -51,6 +51,7 @@ data:
     health-check-probe-interval: 1000000000
     health-check-reply-timeout: 500000000
     health-check-threshold: 3
+    reply-timeout: 3000000000
```
lukasmacko: please add the option to the template as well https://github.com/contiv/vpp/blob/master/k8s/contiv-vpp/templates/vpp.yaml#L88

PR #763, opened 2018-05-01
Allow using k8s persistent volumes for etcd

nbartos:


ccrouch:
It would be fantastic if this could merge soon

PR #778, opened 2018-05-11
Address comments in contiv-vpp-bug-report.sh

jmedved:


jmedved:
superceded by nbartos's work

PR #791, opened 2018-05-16
Bug fix for pod match expressions

brecode:
- This PR fixes a bug where values of "IN" and "NOT_IN" key operators
where AND'ed where they should be OR'ed. Also fixes a bug where over
iterations matching pods should be AND'ed and not OR'ed

- Added unit tests with multiple values to test the above operators.


milanlenco:
Thanks, it fixes all known issues in match_expression.go.
However, the newly added test is still failing - quickly going through cache_impl.go I could find two more bugs: LookupPodsByLabelSelector() does not include match expressions and LookupPoliciesByPod() is just all wrong.

I'm merging this one and will try to prepare fixes for the aforementioned issues, but code quality is something we really need to discuss...

PR #829, opened 2018-06-01
Allow setting updateStrategy as helm value

nbartos:
According to
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

"The RollingUpdate update strategy implements automated, rolling
update for the Pods in a StatefulSet. It is the default strategy
when .spec.updateStrategy is left unspecified."

So, this change will not introduce any behavior change unless the
defaults of the new values are overwritten.

### DIFF
--- a/k8s/contiv-vpp.yaml
+++ b/k8s/contiv-vpp.yaml
@@ -145,6 +145,8 @@ spec:
   # Match contiv-etcd DaemonSet.
   selector:
     k8s-app: contiv-etcd
+  updateStrategy:
+    type: RollingUpdate
   ports:
   - port: 12379
     nodePort: 32379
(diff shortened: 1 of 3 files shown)


rastislavszabo:
Hi,

this PR broke the default deployment:

```
k8s-master: error: error validating "/tmp/contiv-vpp.yaml": error validating data: ValidationError(Service.spec): unknown field "updateStrategy" in io.k8s.api.core.v1.ServiceSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

(updateStrategy cannot be specified for a service)

I will prepare a fix.

rastislavszabo:
Fixed in https://github.com/contiv/vpp/pull/831, tests are passing again.

PR #853, opened 2018-06-08
Added support for contiv-vpp image tag

brecode:
- Decommissioned vmware support
- Added support for release branch testing

To test with a different release tag, set the CONTIV_IMAGE_TAG variable in the vagrant-up script. 
(i.e - from latest -> v1.0-179-gb853346-a9fb77d)


rastislavszabo:
Hi @brecode, it seems that since we merged this, 3 CI tests relying on internet connectivity started failing. Could you please check?

PR #902, opened 2018-06-21
Enhance contiv-vpp-bug-report.sh to generate container NW report

rastislavszabo:
This MR extends the the `contiv-vpp-bug-report.sh` script to generate container NW report containing Linux interfaces, routes and ARP table dump for each running Docker container, e.g.:

```
Container k8s_POD_kube-dns-6f4fd4bdf-86cwd_kube-system_fc638360-748a-11e8-aedc-080027de08ea_0
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
77: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UNKNOWN group default qlen 1000
    link/ether 00:00:00:00:00:02 brd ff:ff:ff:ff:ff:ff
    inet 10.1.1.2/32 brd 10.1.1.2 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::bca4:51ff:fe01:956c/64 scope link 
       valid_lft forever preferred_lft forever
default via 10.1.1.1 dev eth0 
10.1.1.1 dev eth0  scope link 
? (10.1.1.1) at 02:fe:fc:07:21:82 [ether] PERM on eth0
```

### DIFF
--- a/scripts/contiv-vpp-bug-report.sh
+++ b/scripts/contiv-vpp-bug-report.sh
@@ -157,7 +157,7 @@ save_container_nw_report() {
         if [[ $ADDR = *"vpp1:"* ]]; then
             echo "Host networking" >> "${CONTAINER_NW_REPORT_FILE}"
         else
-            echo "$ADDR" >> container-report.txt
+            echo "$ADDR" >> "${CONTAINER_NW_REPORT_FILE}"
             read_node_shell_data "sudo nsenter -t ${PID} -n ip route" >> "${CONTAINER_NW_REPORT_FILE}"
             read_node_shell_data "sudo nsenter -t ${PID} -n arp -na" >> "${CONTAINER_NW_REPORT_FILE}"
         fi
@@ -186,7 +186,7 @@ WARNINGS=1
 TIMESTAMP="$(date '+%Y-%m-%d-%H-%M')"
 REPORT_DIR="contiv-vpp-bug-report-$TIMESTAMP"
 STDERR_LOGFILE="script-stderr.log"
-CONTAINER_NW_REPORT_FILE="container-report.txt"
+CONTAINER_NW_REPORT_FILE="container-nw-report.txt"
 
 # What we want to collect is defined globally in arrays, because sometimes we need to use multiple methods of
 # collection.  The array key is the name of the log file for the given command.
--- a/scripts/contiv-vpp-bug-report.sh
+++ b/scripts/contiv-vpp-bug-report.sh
@@ -115,9 +115,9 @@ read_shell_data_ssh() {
 read_node_shell_data() {
     if [ "$USE_SSH" = "1" ]
     then
-        read_shell_data_ssh $@
+        read_shell_data_ssh "$@"
     else
-        read_shell_data_local $@
+        read_shell_data_local "$@"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
 }
 
+save_container_nw_report() {
+    # generates Docker container network report for the $NODE_NAME into the $CONTAINER_NW_REPORT_FILE
+    # for each docker container, prints out its interfaces, routing table and ARP table
+    echo " - ${CONTAINER_NW_REPORT_FILE}"
+
+    # get ID + NAME of each container except the pause containers
+    CONTAINERS_TXT=$(read_node_shell_data "sudo docker ps --format '{{.ID}} {{.Names}}' | grep -v pause")
+    while read -r line; do CONTAINERS+=( "$line" ); done <<< "$CONTAINERS_TXT"
```
nbartos: To be consistent with the other lines in this file, please split it out like so.  Also, please use lower case for local function variables.
```
containers=()
while read -r line
do
    containers+=( "$line" )
done <<< "$containers_txt"
```
The top declaration (which could also be written 'declare -a containers') is needed in case there are no matching lines, otherwise due to the 'set -u' at the top of the file, trying to use the containers variable below will blow up with an uninitialized variable error.

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+
+    # return in case of no container data (e.g. issues with executing sudo)
+    if [ -z "${CONTAINERS}" ]; then
```
nbartos: To keep the style consistent, please do split out the 'then' into a separate line.  Also, when checking if there are array members, it's better to get the array element count:
```
if [ "${#containers[@]}" = 0 ]
then
```

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+        return
+    fi
+
+    # loop over each container
```
nbartos: This comment doesn't really serve a point.  Usually comments should be reserved for:
1.  If it's not immediately evident why you are doing something.
2.  If it's not immediately evident what you are doing.

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+    for CONTAINER_DATA in "${CONTAINERS[@]}"
```
nbartos: Please use lower case for local variables.

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+    do
+        # split CONTAINER_DATA to container ID and container Name
+        IFS=' ' read -ra CINFO <<< "$CONTAINER_DATA"
+
+        echo "" >> "${CONTAINER_NW_REPORT_FILE}"
```
nbartos: You don't need the "" in the echo.  To keep the style consistent with the rest of the file, please don't use {} unless you need to, e.g.:
```
echo >> "$CONTAINER_NW_REPORT_FILE"

File: scripts/contiv-vpp-bug-report.sh [hunk 6]
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+    # return in case of no container data (e.g. issues with executing sudo)
+    if [ -z "${CONTAINERS}" ]; then
+        return
+    fi
+
+    # loop over each container
+    for CONTAINER_DATA in "${CONTAINERS[@]}"
+    do
+        # split CONTAINER_DATA to container ID and container Name
+        IFS=' ' read -ra CINFO <<< "$CONTAINER_DATA"
```
nbartos: Should use lower case local variables.  Also, it might be better to use individual variables here instead of reading into an array.  That way if there are less elements than you expect, you won't get an undefined variable error below.

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+
+        echo "" >> "${CONTAINER_NW_REPORT_FILE}"
+        echo "Container ${CINFO[1]}" >> "${CONTAINER_NW_REPORT_FILE}"
+
+        PID=$(read_node_shell_data "sudo docker inspect --format '{{.State.Pid}}' ${CINFO[0]}")
```
nbartos: Shouldn't use "" over the whole arg, e.g.:
```
pid=$(read_node_shell_data sudo docker inspect --format '{{.State.Pid}}' "${cinfo[0]}")
```

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+        ADDR=$(read_node_shell_data "sudo nsenter -t ${PID} -n ip addr")
```
nbartos: addr=$(read_node_shell_data sudo nsenter -t "$pid" -n ip addr)

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+
+        if [[ $ADDR = *"vpp1:"* ]]; then
```
nbartos: please put then on second line

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
+            echo "Host networking" >> "${CONTAINER_NW_REPORT_FILE}"
+        else
+            echo "$ADDR" >> "${CONTAINER_NW_REPORT_FILE}"
+            read_node_shell_data "sudo nsenter -t ${PID} -n ip route" >> "${CONTAINER_NW_REPORT_FILE}"
```
nbartos: Same with arg quoting as above

scripts/contiv-vpp-bug-report.sh
```diff
@@ -329,6 +384,9 @@ then
                 get_shell_data_ssh "$CMD_INDEX" ${LOCAL_COMMANDS[$CMD_INDEX]} </dev/null
             done
 
+            # save container network report for this node
```
nbartos: This comment isn't really useful.

scripts/contiv-vpp-bug-report.sh
```diff
@@ -104,13 +104,66 @@ get_shell_data_ssh() {
...
     master_kubectl "$@"> "$log"
 }
 
+save_container_nw_report() {
+    # generates Docker container network report for the $NODE_NAME into the $CONTAINER_NW_REPORT_FILE
+    # for each docker container, prints out its interfaces, routing table and ARP table
+    echo " - ${CONTAINER_NW_REPORT_FILE}"
+
+    # get ID + NAME of each container except the pause containers
+    CONTAINERS_TXT=$(read_node_shell_data "sudo docker ps --format '{{.ID}} {{.Names}}' | grep -v pause")
```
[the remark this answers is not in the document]
rastislavszabo: It actually does not work without double-quoting this. I guess because of the `|`? Anyway, does not work even with escaping the pipe, so leaving the quotes there.

scripts/contiv-vpp-bug-report.sh
rastislavszabo: The above loop actually produces 1 array item in case of empty `$containers_txt` string. Moving the check above the array construction.

rastislavszabo:
Code review issues addressed & retested.

PR #929, opened 2018-07-04
Cleanup shellcheck errors/warnings

jmedved:
- Cleanup shellcheck errors/warnings
- Fix a problem with printing errors in ARP Table check
- Gracefully handle situations where IP & MAC addresses are missing from the vpp address log files


scripts/contiv-error-report.sh
```diff
@@ -21,27 +21,25 @@ usage() {
...
-    for x in "$@"
+get_log_file_name() {
+    for file in "$1"/{.*,*}
     do
-       # echo "$x" >&2
-       if [ "${#x}" -gt "$m" ]
-       then
-          m="${#x}"
-       fi
+        LOG_FILE_NAME=$( echo "$file" | grep "$2" ) || true
```
nbartos: You probably want something like:
```
if LOG_FILE_NAME=$( echo "$file" | grep "$2" ) && [ -n "$LOG_FILE_NAME" ]
```

scripts/contiv-error-report.sh
```diff
@@ -67,7 +65,7 @@ done
 pushd "$REPORT_DIR" > /dev/null
 
 # Get all nodes in the cluster and their host IP addresses
-NODES=$(cat "$NODES_FILE" | grep -v "NAME")
+NODES=$( grep < "$NODES_FILE" -v "NAME")
```
nbartos: The grep arguments should come before the redirect, e.g.:
```
NODES=$( grep -v "NAME" < "$NODES_FILE")
```
Also, grepping without a delimiter isn't really a good idea.  You might end up filtering out things you wouldn't want.  I'm not sure what the output looks like exactly, but something like this would be better:

If the whole line consists of just "NAME":
```
NODES=$( grep -vxF "NAME" < "$NODES_FILE")
```

If the line starts with "NAME" and then is followed by a space and something else:
```
NODES=$( grep -v ^$NAME " < "$NODES_FILE")
```

scripts/contiv-error-report.sh
```diff
@@ -76,35 +74,32 @@ do
...
 for nn in "${NODE_NAMES[@]}"
 do
     # Print node header
-    printf "%s:\n" "$nn"
+    printf "%s:\\n" "$nn"
     printf '%0.s=' $( seq 1 $(( ${#nn} + 1 )) )
     echo
 
-    VSWITCH_LOG_FILE_NAME=$( ls "$nn" | grep "$VSWITCH_LOG" )
+    VSWITCH_LOG_FILE_NAME=$( get_log_file_name "$nn" "$VSWITCH_LOG" )
```
nbartos: This seems to be doing a grep without using a delimiter, and possibly VSWITCH_LOG may contain special regex characters that might need to be escaped.

scripts/contiv-error-report.sh
```diff
@@ -76,35 +74,32 @@ do
...
     if [ -n "$VSWITCH_LOG_FILE_NAME" ]
     then
         echo "Vswitch log:"
         echo "------------"
-        cat "$nn"/"$VSWITCH_LOG_FILE_NAME" | grep -n "$SEARCH_STRING" || true
+        grep < "$VSWITCH_LOG_FILE_NAME" -n "$SEARCH_STRING" || true
```
nbartos: Redirect should be after grep arguments.

scripts/contiv-error-report.sh
```diff
@@ -76,35 +74,32 @@ do
...
     if [ "$SHOW_PREVIOUS" == "1" ]
     then
-        VSWITCH_LOG_FILE_NAME=$( ls "$nn" | grep "$VSWITCH_PREVIOUS_LOG" ) || true
+        VSWITCH_LOG_FILE_NAME=$( get_log_file_name "$nn" "$VSWITCH_PREVIOUS_LOG" )
         if [ -n "$VSWITCH_LOG_FILE_NAME" ]
         then
             echo "Previous vswitch log:"
             echo "---------------------"
-            cat "$nn"/"$VSWITCH_LOG_FILE_NAME" | grep -n "$SEARCH_STRING"
+            grep < "$VSWITCH_LOG_FILE_NAME" -n "$SEARCH_STRING"
```
nbartos: Redirect should be after grep arguments.

scripts/contiv-error-report.sh
```diff
@@ -76,35 +74,32 @@ do
...
+        grep < "$VSWITCH_LOG_FILE_NAME" -n "$SEARCH_STRING" || true
     else
         echo "Logfile for contiv-vswitch not present."
     fi
     echo
 
     if [ "$SHOW_PREVIOUS" == "1" ]
     then
-        VSWITCH_LOG_FILE_NAME=$( ls "$nn" | grep "$VSWITCH_PREVIOUS_LOG" ) || true
+        VSWITCH_LOG_FILE_NAME=$( get_log_file_name "$nn" "$VSWITCH_PREVIOUS_LOG" )
```
nbartos: Seems to be grepping without delimiters.

scripts/contiv-network-report.sh
```diff
@@ -123,7 +121,7 @@ done
 pushd "$REPORT_DIR" > /dev/null
 
 # Get all nodes in the cluster and their host IP addresses
-NODES=$( cat "$NODES_FILE" | grep -v "NAME" ) || true
+NODES=$( grep < "$NODES_FILE" -v "NAME" ) || true
```
nbartos: Redirect should be after grep args.

scripts/contiv-network-report.sh
```diff
@@ -205,18 +202,18 @@ do
...
-        elif [ $( trim "${IF_NAME_STATUS[0]}" ) == "L3" ]
+        elif [ "$( trim "${IF_NAME_STATUS[0]}" )" == "L3" ]
         then
             IF_IP["$IF_NAME"]=$( trim "${IF_NAME_STATUS[1]}" )
         fi
     done
 
     # Get MAC addresses for all interfaces that have an IP address
-    VPP_MAC_ADDR=$( cat "$nn"/"$VPP_MAC_ADDR_FILE" | grep -v "Name" ) || true
+    VPP_MAC_ADDR=$( grep < "$nn"/"$VPP_MAC_ADDR_FILE" -v "Name" ) || true
```
nbartos: Redirect should be after grep args.  Also grep expression should have delimiters or be full line match.

scripts/contiv-network-report.sh
```diff
@@ -239,6 +236,19 @@ do
         fi
     done
 
+    # Handle cases where IP or MAC address was not present in the vpp log
+    for ifn in "${IF_NAMES[@]}"
+    do
+        if [ ! ${IF_IP[$ifn]+_} ]
```
nbartos: I think ${IF_IP[$ifn]+_} should be quoted.

scripts/contiv-network-report.sh
```diff
@@ -239,6 +236,19 @@ do
...
+        then
+            IF_IP[$ifn]="<none>"
+        fi
+        if [ ! ${IF_MAC[$ifn]+_} ]
```
nbartos: ${IF_MAC[$ifn]+_} should probably be quoted.

scripts/contiv-network-report.sh
```diff
@@ -347,19 +356,18 @@ then
...
-                ERROR_LINE=$( echo "No VXLANs configured for the node" )
+                ERROR_LINE="No VXLANs configured for the node"
                 ERROR_LINES+=("$ERROR_LINE")
             else
                 VXLAN_MAP["$IF_IDX"]="$l"
             fi
         done
 
-        L2FIB=$( cat "$nn/$VPP_L2FIB_FILE" | grep -v "Mac-Address" | grep -v "L2FIB" ) || true
+        L2FIB=$( grep < "$nn/$VPP_L2FIB_FILE" -v "Mac-Address" | grep -v "L2FIB" ) || true
```
nbartos: Redirect should be after grep args.  Grep expressions should use delimiters.

scripts/contiv-error-report.sh
```diff
@@ -21,27 +21,24 @@ usage() {
...
-    for x in "$@"
+get_log_file_name() {
+    for file in "$1"/{.*,*}
     do
-       # echo "$x" >&2
-       if [ "${#x}" -gt "$m" ]
-       then
-          m="${#x}"
-       fi
+        if LOG_FILE_NAME=$( echo "$file" | grep "$2" ) && [ -n "$LOG_FILE_NAME" ]
```
nbartos: This is a local variable, so it should be lower case.

scripts/contiv-network-report.sh
```diff
@@ -205,18 +202,18 @@ do
...
-        elif [ $( trim "${IF_NAME_STATUS[0]}" ) == "L3" ]
+        elif [ "$( trim "${IF_NAME_STATUS[0]}" )" == "L3" ]
         then
             IF_IP["$IF_NAME"]=$( trim "${IF_NAME_STATUS[1]}" )
         fi
     done
 
     # Get MAC addresses for all interfaces that have an IP address
-    VPP_MAC_ADDR=$( cat "$nn"/"$VPP_MAC_ADDR_FILE" | grep -v "Name" ) || true
+    VPP_MAC_ADDR=$( grep -v "Name " < "$nn"/"$VPP_MAC_ADDR_FILE" ) || true
```
nbartos: Doing a single string quote makes this a bit easier to read:
```
VPP_MAC_ADDR=$( grep -v "Name " < "$nn/$VPP_MAC_ADDR_FILE" ) || true
```

scripts/contiv-network-report.sh
```diff
@@ -205,18 +202,18 @@ do
...
     if [ -z "$VPP_IP_ADDR" ]
     then
         printf "$NODE_ERR_FORMAT" "$MISSING_NODE" ""
-        echo "Missing or empty hardware address log: '$nn"/"$VPP_MAC_ADDR_FILE'"
+        echo "ERROR: Missing or empty hardware address log: \\'""$nn""/$VPP_MAC_ADDR_FILE\\'"
```
nbartos: All of these extra quotes don't seem to be helping, doing this would be simpler:
```
echo "ERROR: Missing or empty hardware address log: $nn/$VPP_MAC_ADDR_FILE"
```
or
```
echo "ERROR: Missing or empty hardware address log: '$nn/$VPP_MAC_ADDR_FILE'"
```

PR #933, opened 2018-07-08
LUNAR #1435 - add script for harvesting core dumps

amarcine:


scripts/contiv-vpp-bug-report.sh
```diff
@@ -221,6 +221,7 @@ LOCAL_COMMANDS["linux-ip-route.log"]="ip route"
 LOCAL_COMMANDS["contiv-stn.log"]='CONTAINER=$(sudo docker ps --filter name=contiv-stn --format "{{.ID}}") && [ -n "$CONTAINER" ] && sudo docker logs "$CONTAINER"'
 LOCAL_COMMANDS["vswitch-version.log"]="curl -m 2 localhost:9999/liveness"
 LOCAL_COMMANDS["docker-ps.log"]="sudo docker ps"
+LOCAL_COMMANDS["core-dump.log"]='mkdir -p /vagrant/dumps && sudo tar -zcf /vagrant/dumps/`hostname`-dump.tgz /run/vpp/dumps && ls -alR /vagrant/dumps'
```
nbartos: This will run on non-vagrant things.  Things that are for vagrant only probably shouldn't be in this script.

rastislavszabo: maybe we could use scp to copy core files from the nodes?

amarcine: it was just folder name, so i change it to /tmp to not be confused by name 'vagrant'

PR #945, opened 2018-07-20
Introduce POD VRF

rastislavszabo:
This MR introduces separate VRF for POD traffic. Going forward, we can support multiple POD VRFs for multiple tenants. The main and POD VRF ID can be changed via YAML, but leaving that as a "secret option" for now.

This MR also enhances STN deplyment scenario - addresses most of the networking issues. There is still one outstanding STN issue related to fragmented packets (https://jira.fd.io/browse/VPP-1325).

Also, there might be an issue in case of services with mixed HostNetwork and non-HostNetwork backends - will be fixed once https://jira.fd.io/browse/VPP-1345 is implemented and integrated.

The MR introduces changes in the vpp-agent in the vendor directory until https://github.com/ligato/vpp-agent/pull/747 is merged and propagated into the contiv repo.


milanlenco:
Would be great to have this covered by service UTs, but otherwise OK.

PR #976, opened 2018-08-02
bug fixes and better error handling.

omegaman123:


plugins/crd/cache/processor.go
```diff
@@ -159,6 +159,7 @@ func (p *ContivTelemetryProcessor) getNodeDTOInfo(client http.Client, node *Node
 	} else if res.StatusCode < 200 || res.StatusCode > 299 {
 		err := fmt.Errorf("getNodeDTOInfo: url: %s HTTP responsres.Status: %s", url, res.Status)
 		p.nodeResponseChannel <- &NodeDTO{node.Name, nil, err}
```
jmedved: Please log an error, just like in the if statement above

PR #977, opened 2018-08-02
Allow vagrant to work behind a proxy

mestery:
After spending a large amount of time trying to get this working behind a
proxy, I was finally able to make it work with the changes in this PR.
These include:

* Use `add-apt-repository` to correctly add apt repositories.
* Use `sudo -E` to ensure we pass any proxy environment variables to
  commands run by sudo.
* Make sure to export HTTP_PROXY and HTTPS_PROXY.
* Setup the docker proxy via systemd configuration if a proxy is set.

With these changes, I was able to bringup contiv vpp with vagrant behind
a proxy.

Signed-off-by: Kyle Mestery <mestery@mestery.com>


jmedved:
Awesome thanks a lot!

PR #995, opened 2018-08-07
 Workaround for the DNS failure: check & wait for POD IP using netlink before returning CNI reply

rastislavszabo:


plugins/contiv/remote_cni_server.go
```diff
@@ -1077,6 +1077,13 @@ func (s *remoteCNIserver) configureContainerConnectivity(request *cni.CNIRequest
 		}
 	}
 
+	// verify that the POD has the allocated IP address configured / wait until it is actually configured
+	err = s.verifyPodIP(request.NetworkNamespace, request.InterfaceName, podIP)
+	if err != nil {
```
lukasmacko: i would suggest extend the condition to `|| s.test` in order to pass unit tests

PR #1002, opened 2018-08-08
Add k8s contiv/vpp Helm support for arm64 platform

TrevorTaoARM:
Add k8s Helm template and makefile support for the generation of the deployment yaml file
of arm64 platform.

Give only different parts in values-arm64.yaml for arm64 from values.yaml.

A sample file named contiv-vpp-arm64.yaml is also given.

### DIFF
--- a/k8s/contiv-vpp/README.md
+++ b/k8s/contiv-vpp/README.md
@@ -24,7 +24,7 @@ kubectl apply -f manifest.yaml
 To install without tiller for arm64, you can generate the manifest from this chart and install via kubectl:
 
 ```console
-helm template --name my-release ../contiv-vpp -f ./values-arm64.yaml > manifest-arm64.yaml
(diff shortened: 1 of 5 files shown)


k8s/contiv-vpp/README.md
```diff
@@ -21,6 +21,13 @@ helm template --name my-release ../contiv-vpp > manifest.yaml
 kubectl apply -f manifest.yaml
 ```
 
+To install without tiller for arm64, you can generate the manifest from this chart and install via kubectl:
+
+```console
+helm template --name my-release ../contiv-vpp -f ./values-arm64.yaml > manifest-arm64.yaml
```
rastislavszabo: shouldn't this contain both: `-f ./values-arm64.yaml,./values.yaml` ?

TrevorTaoARM: Yes, it should be added. 
Updated.

PR #1026, opened 2018-08-14
Documented special case: Kubernetes 1.11 two node cluster on Cavium ThunderX (ARM64) machines

stanislav-chlebec:
Here are documented the differences of deployment on Cavium ThunderX.


stanislav-chlebec:
Will be adjusted according Trevor:
I don’t think the hugepage section in contiv-vpp-arm64.yaml is necessary:
353 +           resources:            
 354 +             limits:  
 355 +               hugepages-1Gi: 8Gi  
 356 +               memory: 8Gi  

You can set it in the kubernetes as that in MANUAL_INSTALL.md:
To disable huge pages, perform the following steps as root:
•	Using your favorite editor, disable huge pages in the kubelet configuration file (/etc/systemd/system/kubelet.service.d/10-kubeadm.conf):
  Environment="KUBELET_EXTRA_ARGS=--feature-gates HugePages=false"

And it works in my environment.

k8s/contiv-vpp-arm64.yaml
```diff
@@ -26,7 +26,7 @@ data:
     UseTAPInterfaces: true
     TAPInterfaceVersion: 2
     NatExternalTraffic: true
-    MTUSize: 1450
+    MTUSize: 1500
```
rastislavszabo: Do not commit changes in yaml file without changing the helm template - it should reflect helm generation with default values

PR #1030, opened 2018-08-14
Added collection of node static routes.

omegaman123:


plugins/crd/cache/telemetry_cache.go
```diff
@@ -42,6 +42,7 @@ const (
 	arpURL             = "/arps"
 	clientTimeout      = 10 // HTTP client timeout, in seconds
 	collectionInterval = 1  // data collection interval, in minutes
+	staticRouteURL     = "/staticroutes"
 )
```
jmedved: Move this up to the other urls - please try to keep the URLs together

PR #1035, opened 2018-08-15
added timestamp for report and versioning of data updates.

omegaman123:


plugins/crd/datastore/report_test.go
```diff
@@ -0,0 +1,33 @@
...
+// limitations under the License.
+
+package datastore
+
+import (
+	"github.com/ligato/cn-infra/logging/logrus"
+	"testing"
+	"time"
+)
+
```
jmedved: Add this mock log writer somewhere here:
// mockLogWriter collects all error logs into a buffer for analysis
// by gomega assertions.
type mockLogWriter struct {
	log []string
}

func (mlw *mockLogWriter) Write(p []byte) (n int, err error) {
	logStr := string(p)
	mlw.log = append(mlw.log, logStr)
	return len(logStr), nil
}

func (mlw *mockLogWriter) clearLog() {
	mlw.log = []string{}
}

func (mlw *mockLogWriter) printLog() {
	fmt.Println("Error log:")
	fmt.Println("==========")
	for i, l := range mlw.log {
		fmt.Printf("%d: %s", i, l)
	}
}

func (mlw *mockLogWriter) countErrors() int {
	cnt := 0
	for _, logLine := range mlw.log {
		if strings.Contains(logLine, "level=error") {
			cnt++
		}
	}
	return cnt
}

plugins/crd/datastore/report_test.go
```diff
@@ -0,0 +1,33 @@
...
+func TestSimpleReport_AppendToNodeReport(t *testing.T) {
+	report := NewSimpleReport(logrus.DefaultLogger())
+	report.LogErrAndAppendToNodeReport("nodeName", "ErrorString")
```
jmedved: do:
lw = = &mockLogWriter{log: []string{}}
report.Log.SetOutput(lw)

plugins/crd/datastore/report_test.go
```diff
@@ -0,0 +1,33 @@
...
+	report.SetTimeStamp(time.Now())
+	time := report.GetTimeStamp()
+	report.SetTimeStamp(time)
+	report.Print()
```
jmedved: check that what you get back is the same that you wrote there.

plugins/crd/datastore/report_test.go
```diff
@@ -0,0 +1,33 @@
...
+package datastore
+
+import (
+	"github.com/ligato/cn-infra/logging/logrus"
+	"testing"
+	"time"
+)
+
+func TestSimpleReport_AppendToNodeReport(t *testing.T) {
+	report := NewSimpleReport(logrus.DefaultLogger())
```
jmedved: here do:
lw = = &mockLogWriter{log: []string{}}
report.Log.SetOutput(lw)

plugins/crd/datastore/report_test.go
jmedved: here check that the length of log in lw is 1 
also, check that length of report.Data[] is 1

PR #1039, opened 2018-08-17
Expose IPAM state via REST

deathbeam:
Expose IPAM state:
```
{
  nodeId,
  nodeIP,
  nodeName,
  podNetwork,
  vppHostNetwork
}
```

via HTPP REST interface (endpoint is `GET /ipam`).

Example output:

```json
$ http 10.0.2.15:9999/ipam
HTTP/1.1 200 OK
Content-Length: 158
Content-Type: application/json; charset=UTF-8
Date: Fri, 17 Aug 2018 11:07:53 GMT

{
    "nodeIP": "192.168.16.1",
    "nodeId": 1,
    "nodeName": "vagrant-arch.vagrantup.com",
    "podNetwork": "10.1.1.0/24",
    "vppHostNetwork": "172.30.1.0/24"
}
```

Signed-off-by: Tomas Slusny <slusnucky@gmail.com>

### DIFF
--- a/plugins/contiv/ipam/ipam.go
+++ b/plugins/contiv/ipam/ipam.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/ligato/cn-infra/db/keyval"
 	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/rpc/rest"
 	"sort"
 )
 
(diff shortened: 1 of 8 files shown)


jmedved:
This is a good start, but the URL should be versioned, i.e. not just ipam, but something like contiv/v1/ipam. We have those kinds of URLs for vpp dumps, we should keep the same structure. Please talk to Ondrej Fabry about what exactly it should be.

PR #1078, opened 2018-08-30
Contiv CRD: Propagate NodeConfigs to ETCD - without resync.

milanlenco:
In order to test the agent's side of CRD-based node config I needed to implement the basic propagation of NodeConfig specs from contiv-crd to ETCD.
Nikos, feel free to re-use / re-work the implementation. Also, find inspiration in the resync implementation from KSR's markAndSweep() procedure in plugins/ksr/ksr_reflector.go (potentially expose the method and re-use to avoid code duplication).


plugins/crd/controller/nodeconfig/node_config_controller.go
```diff
@@ -120,21 +123,27 @@ func (c *Controller) Init() error {
 			newEvent.resource = new
```
brecode: Let's change it to oldObj, newObj so it won't be confused with keyword "new"

brecode: Let's change it to oldObj, newObj so it won't be confused with keyword "new"

brecode:
Merging but please see comments -good stuff ! 👍

plugins/crd/controller/nodeconfig/node_config_controller.go
```diff
@@ -210,35 +219,23 @@ func (c *Controller) processNextItem() bool {
...
 			return nil
 		}
 	case "update":
-		updateEvent := Event{
-			key:         newEvent.key,
-			oldResource: newEvent.oldResource,
-			resource:    newEvent.resource,
-		}
-		c.eventHandler.ObjectUpdated(updateEvent)
+		c.eventHandler.ObjectUpdated(event.resource)
```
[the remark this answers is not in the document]
milanlenco: You are probably right - I have also noticed that resources contain timestamp which is changing all the time and that causes the ObjectUpdated to be called regularly (at least I think that is the reason). To reduce number of writes into etcd, a check for equality of propagated attributes is probably going to be needed,

plugins/crd/handler/nodeconfig/node_config_handler.go
```diff
@@ -28,16 +45,65 @@ func (h *Handler) Init() error {
 
 // ObjectCreated is called when a CRD object is created
 func (h *Handler) ObjectCreated(obj interface{}) {
-	//fmt.Printf("Object created with value: %v", obj)
+	h.Log.Debugf("Object created with value: %v", obj)
+	nodeConfig, ok := obj.(*v1.NodeConfig)
```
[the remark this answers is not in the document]
milanlenco: I think not, nodeConfig is immediately converted and copied to protobuf structure. The same holds for ObjectDeleted() and ObjectUpdated().

plugins/crd/controller/nodeconfig/node_config_controller.go
milanlenco: agree

PR #1101, opened 2018-09-10
[WIP] contiv-crd: Updated build process to include Contiv-CRD

brecode:
* Updated Makefile to include build for contiv-crd
  * Updated Dockerfiles to include contiv-crd docker image build
  * Created model for crd plugin status to be reflected into ETCD
  * Push information on etcd on CRD plugin AfterInit()
  * Updated k8s/contiv-vpp.yaml to include the Contiv-CRD agent
    as part of the Contiv-VPP CNI plugin

Signed-off-by: Nikos Bregiannis <nikos.bre@gmail.com>


docker/ubuntu-based/vpp/Dockerfile
```diff
@@ -17,6 +17,16 @@ RUN wget -O go.tgz "https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz"
...
+# install ETCD
+ENV ETCD_VER=v3.3.9
+ENV GOOGLE_URL=https://storage.googleapis.com/etcd
+ENV GITHUB_URL=https://github.com/coreos/etcd/releases/download
+ENV DOWNLOAD_URL=${GOOGLE_URL}
+
+RUN curl -L ${DOWNLOAD_URL}/${ETCD_VER}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
+RUN tar xzvf /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz -C /tmp/etcd --strip-components=1
+RUN rm -f /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
+
```
milanlenco: Installation of ETCD should not be part of this image build process - it's purpose is to build VPP only.

docker/ubuntu-based/prod/vswitch/Dockerfile
```diff
@@ -13,6 +13,16 @@ RUN apt-get update \
...
  && rm -rf /var/lib/apt/lists/*
 
+# install ETCD
+ENV ETCD_VER=v3.3.9
+ENV GOOGLE_URL=https://storage.googleapis.com/etcd
+ENV GITHUB_URL=https://github.com/coreos/etcd/releases/download
+ENV DOWNLOAD_URL=${GOOGLE_URL}
+
+RUN curl -L ${DOWNLOAD_URL}/${ETCD_VER}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
+RUN tar xzvf /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz -C /tmp/etcd --strip-components=1
```
milanlenco: Put all RUN commands into one so that intermediate layers are not produced.

milanlenco: We do we need to install ETCD anyway? My understanding is that we are using etcdctl for the init container from quay.io/coreos/etcd:v3.1.10

k8s/contiv-vpp.yaml
```diff
@@ -254,22 +254,34 @@ spec:
...
+        - name: ETCDCTL_API
+          value: "3"
+        image: quay.io/coreos/etcd:v3.1.10
+        command:
+        - /bin/sh
+        args:
+        - -c
+        - |
+          set -eu
+          until /usr/local/bin/etcdctl --endpoints=127.0.0.1:$ETCDPORT get "/vnf-agent/contiv-ksr/status/crd" --prefix=true | grep -m 1 "1";
```
milanlenco: I will need to talk about this startup dependency with Lukas and Rasto once they are back from PTO. At this point I'm not sure if it is really needed + we want to be able to start node independently on the accessibility of ETCD eventually.

docker/ubuntu-based/vpp/Dockerfile
brecode: Leftovers from testing, shouldn't be there since I moved to the quay.io/coreos/etcd:v3.1.10 removing....

docker/ubuntu-based/prod/vswitch/Dockerfile
brecode: Leftovers from testing, shouldn't be there since I moved to the quay.io/coreos/etcd:v3.1.10 removing....

k8s/contiv-vpp.yaml
brecode: I don't mind adding a mechanism in the yaml to check if contiv-crd agent is up and running, but we will need to find a way to resolve this as I think having CRD and KSR depending on ETCD and then other components depend on CRD/KSR readiness will probably cause problems in the future.

plugins/contiv/config.go
```diff
@@ -0,0 +1,181 @@
...
+	// try remote kv-store first
+	if remoteDB != nil {
+		nodeConfigProto, err = loadNodeConfigFromKVStore(nodeName, remoteDB)
+		if err != nil {
+			log.WithField("err", err).Warn("Failed to read node configuration from remote KV-store")
+		}
+		if err == nil {
+			// mirror the config into the local kv-store
+			// TODO: remove once all kubernetes state data are reflected into local KV-store by the Aggregator
+			boltBroker := localDB.NewBroker(servicelabel.GetDifferentAgentPrefix(ksr.MicroserviceLabel))
```
lukasmacko: add nil check for localDB

k8s/contiv-vpp.yaml
rastislavszabo: Currently, the vswitch agent needs to wait for ETCD anyway, but as Milan mentioned, we want to make it independent on ETCD later on. For now, we could even go with a wait for CRD data in ETCD, but we still need to solve the startup dependency anyway: since the node config can come to CRD after CRD is started, how do we know if we should wait for it or not? I think we need some kind of a global switch - either CRD or YAML + auto-calculated defaults can be used for node config. We cannot support both at the same time. In case of CRD, the node config would be mandatory even if it is empty. Then the vswitch should always wait until its node config is available in CRD.

plugins/crd/validator/l3/l3_validator.go
```diff
@@ -124,13 +156,17 @@ func (v *Validator) validateVrf1PodRoutes(node *telemetrymodel.Node, vrfMap map[
 			continue
 		}
 
+		// Assume that the route will be valid. Each failed check flips
+		// the status
+		routeMap[1][lookUpRoute.Ipr.DstAddr] = routeValid
```
lukasmacko: I think it would be better to use var or const instead of magic number.

jmedved: routeValid is a constant - or did you want something else?

k8s/contiv-vpp.yaml
jmedved: If you have config data locally cached in Bolt, you can come up and resync with CRD later (i.e. if CRD config for you is different from the cached config, you have no choice but to restart with the new config). If there is no locally cached config, you have to wait for CRD. 

Now, if CRD is initially empty and we do want to keep supporting the initial config though the YAML file, we can have an init container that gets config from the YAML file and shoves it into CRD. We just have to decide what takes precedence in case of conflict - YAML or CRD.

k8s/contiv-vpp/templates/vpp.yaml
```diff
@@ -453,6 +458,27 @@ spec:
...
+        - name: ETCDCTL_API
+          value: "3"
+        image: quay.io/coreos/etcd:v3.1.10
+        command:
+        - /bin/sh
+        args:
+        - -c
+        - |
+          set -eu
+          until /usr/local/bin/etcdctl --endpoints=127.0.0.1:$ETCDPORT get "/vnf-agent/contiv-ksr/k8s/nodeconfig/$HOSTNAME" --prefix=true | grep -m 1 "$HOSTNAME";
```
lukasmacko: if security is enabled we have pass certificates to etcdctl

PR #1110, opened 2018-09-19
ipam: introducing contivCIDR to calculate pod/node/vxlan address spaces

brecode:
This PR removes the ability from the user to define pod/node/vxlan address space.

  * Introducing ContivCIDR in the contiv-vpp.yaml from which, all the subnets are being calculated
    A /14 prefix (or less, i.e /13, /12) is needed. 
  * From the given CIDR we get two subnets of 32768 (or more) addresses each assigned for 
    PodSubnetCIDR and VPPHostSubnetCIDR
  * Three /23 subnets are reserved for NodeInterconnectCIDR, VxlanCIDR and podIfIPCIDR
  * Modified Vagrant to reflect changes 
  * Modified Helm charts
  * Modified yaml (k8s/contiv-vpp + arm)

### DIFF
--- a/k8s/contiv-vpp/values.yaml
+++ b/k8s/contiv-vpp/values.yaml
@@ -14,7 +14,7 @@ contiv:
   serviceLocalEndpointWeight: 1
   disableNATVirtualReassembly: True
   ipamConfig:
-    contivCIDR: "10.1.0.0/15"
+    contivCIDR: "10.0.0.0/14"
     #serviceCIDR: "10.96.0.0/12"
   # example of node configuration for VPP interfaces
   #nodeConfig:
--- a/k8s/contiv-vpp/values.yaml
+++ b/k8s/contiv-vpp/values.yaml
@@ -14,7 +14,7 @@ contiv:
   serviceLocalEndpointWeight: 1
   disableNATVirtualReassembly: True
   ipamConfig:
-    contivCIDR: "10.1.0.0/16"
+    contivCIDR: "10.1.0.0/15"
     #serviceCIDR: "10.96.0.0/12"
   # example of node configuration for VPP interfaces
   #nodeConfig:
--- a/k8s/contiv-vpp/values.yaml
+++ b/k8s/contiv-vpp/values.yaml
@@ -14,13 +14,7 @@ contiv:
   serviceLocalEndpointWeight: 1
   disableNATVirtualReassembly: True
   ipamConfig:
-    podSubnetCIDR: "10.1.0.0/16"
-    podNetworkPrefixLen: 24
-    podIfIPCIDR: 10.2.1.0/24
-    vppHostSubnetCIDR: "172.30.0.0/16"
-    vppHostNetworkPrefixLen: 24
-    vxlanCIDR: "192.168.30.0/24"
-    nodeInterconnectCIDR: "192.168.16.0/24"
+    contivCIDR: "10.1.0.0/16"
     #serviceCIDR: "10.96.0.0/12"
   # example of node configuration for VPP interfaces
   #nodeConfig:
--- a/plugins/contiv/plugin_impl_contiv.go
+++ b/plugins/contiv/plugin_impl_contiv.go
@@ -628,15 +628,15 @@ func appendIfMissing(slice []string, s string) []string {
 func getIPAMConfig(config *Config) *Config {
 	_, contivNetwork, _ := net.ParseCIDR(config.IPAMConfig.ContivCIDR)
 	maskSize, _ := contivNetwork.Mask.Size()
-	subnetPrefixLength := 24 - maskSize
+	subnetPrefixLength := 23 - maskSize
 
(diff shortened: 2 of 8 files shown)


jmedved:
"Three /24 subnets are reserved for NodeInterconnectCIDR, VxlanCIDR and podIfIPCIDR" - this will mean we can only have up to 256 nodes. we need a <=23 subnets for all these to support 500 nodes.

brecode:
Yup, before I change it to /24 I wanted to consult with the @rastislavszabo and @lukasmacko if the nodeID handles more than 256.

plugins/contiv/plugin_impl_contiv.go
```diff
@@ -620,3 +623,100 @@ func appendIfMissing(slice []string, s string) []string {
...
+	return config
+}
+
+// subnet takes a CIDR range and creates a subnet from it
+// base: parent CIDR range
+// newBits: number of additional prefix bits
+// num: given network number.
+//
+// Example: 10.1.0.0/16, with additional 8 bits and a network number of 5
+// result = 10.3.5.0/24
```
lukasmacko: result should be `10.1.5.0/24` shoudln't it?

brecode: yup typo

lukasmacko:
Could you add a simle unit test that will show how the default ContivCIDR is split?

Correct me if I'm wrong however it seems to me that there is an overlap in generated subnets:
Given `10.1.0.0/16` as ContivCIDR

| name   | subnet    |
|---|---|
| podSubnetCIDR  | 10.1.0.0/17 | 
| vppHostSubnetCIDR  | **10.1.128.0/17** includes all three below |
| nodeInterconnectCIDR  | 10.1.250.0/23 |
| podIfIPCIDR  | 10.1.252.0/23 |
| vxlanCIDR  | 10.1.254.0/23 |

brecode:
@lukasmacko You're right, I thought we could handle address overlapping although it's not good practise, so we don't go into a bigger network and have so many unused addresses. 

I've now increased the ContivCIDR to /14 so we can have ~ 65K pods in the cluster and 512 nodes but we're left with a many unused addresses.

plugins/contiv/plugin_impl_contiv.go
```diff
@@ -626,17 +626,17 @@ func appendIfMissing(slice []string, s string) []string {
 
 // getIPAMConfig populates the Config struct with the calculated subnets
 func getIPAMConfig(config *Config) *Config {
-	_, contivNetwork, _ := net.ParseCIDR(config.IPAMConfig.ContivCIDR)
+	_, contivNetwork, _ := net.ParseCIDR("10.1.0.0/14")
```
lukasmacko: the param should be used instead of constant

jmedved:
One more thing occurred to me: we should provide an option to configure NodeInterconnectCIDR separately from the rest of the IP address ranges, since a user may want to allocate their own visible address range for that (e.g. from their own public address space)

brecode:
@jmedved so are we leaving the nodeInterconnectCIDR for the user to decide ?

k8s/contiv-vpp/values.yaml
```diff
@@ -14,7 +14,7 @@ contiv:
   serviceLocalEndpointWeight: 1
   disableNATVirtualReassembly: True
   ipamConfig:
-    contivCIDR: "10.1.0.0/16"
+    contivCIDR: "10.1.0.0/15"
```
lukasmacko: `.1` the second octet will be ingnored due to mask used, rather use `10.2.0.0/15`

plugins/contiv/plugin_impl_contiv.go
brecode: Right, this went in from my testing.

k8s/contiv-vpp/values.yaml
brecode: I think this needs to change to 10.0.0.0/14 now, right?

rastislavszabo:
> One more thing occurred to me: we should provide an option to configure NodeInterconnectCIDR separately from the rest of the IP address ranges, since a user may want to allocate their own visible address range for that (e.g. from their own public address space)

+1 for this

I was even thinking that we might want to still allow to setup all individual subsets by the user as before (instead of ContivCIDR), at least as a "secret" option. Just in case that someone would come in later and tell us they cannot afford so large subnet.

k8s/contiv-vpp/values.yaml
lukasmacko: yes that's another option

brecode:
+1 from me, we actually encountered something similar w/ @jmedved at a debugging session today. We can keep the NodeInterconnectCIDR field and either use the large subnet if it's empty or split the subnets differently if exists in the contiv-vpp.yaml (default value=nil, if introduced by the user we use that instead)

brecode:
Closing this, implemented in #1101

PR #1180, opened 2018-10-15
Refactor interactions with etcd

jmedved:
Connecting to etcd is only done once, and only one connection to etcd will be established.

### DIFF
--- a/plugins/netctl/cmdimpl/cmd_utils.go
+++ b/plugins/netctl/cmdimpl/cmd_utils.go
@@ -36,6 +36,7 @@ type clusterNodeInfo map[string]*node.NodeInfo
 var (
 	nodeInfo    clusterNodeInfo
 	bytesBroker *etcd.BytesConnectionEtcd
+	ipAddrRe    = regexp.MustCompile(`(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}`)
 )
 
 func getEtcdBroker() *etcd.BytesConnectionEtcd {
@@ -64,7 +65,7 @@ func getEtcdBroker() *etcd.BytesConnectionEtcd {
 }
 
 func getClusterNodeInfo() clusterNodeInfo {
-	if (nodeInfo == nil) || (len(nodeInfo) == 0) {
+	if len(nodeInfo) == 0 {
 		db := getEtcdBroker()
 
 		nodeInfo = make(clusterNodeInfo, 0)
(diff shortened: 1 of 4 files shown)


plugins/netctl/cmdimpl/cmd_utils.go
```diff
@@ -31,55 +31,82 @@ import (
...
-	return ""
+
+	return nodeInfo
 }
 
 //resolveNodeOrIP will take in an input string which is either a node name
 // or string and return the ip for the nodename or simply return the ip
-func resolveNodeOrIP(input string) (ipAdr string) {
+func resolveNodeOrIP(nodeName string) (ipAdr string) {
 	re := regexp.MustCompile(`(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}`)
```
ondrej-fabry: This can be defined once as a global variable to avoid unnecessary compilation for each input.

plugins/netctl/cmdimpl/cmd_utils.go
```diff
@@ -31,55 +31,82 @@ import (
...
-	for {
-		kv, stop := itr.GetNext()
-		if stop {
-			break
+
+	return bytesBroker
+}
+
+func getClusterNodeInfo() clusterNodeInfo {
+	if (nodeInfo == nil) || (len(nodeInfo) == 0) {
```
ondrej-fabry: No need to make extra check for nil, `len(nodeInfo)` will return 0 even when it's nil.

plugins/netctl/cmdimpl/nodes.go
```diff
@@ -35,45 +31,17 @@ import (
 
 //PrintNodes will print out all of the cmdimpl in a network in a table format.
 func PrintNodes() {
-	cfg := &etcd.ClientConfig{
-		Config: &clientv3.Config{
-			Endpoints: []string{etcdLocation},
-		},
-		OpTimeout: 1 * time.Second,
+	nodes := make([]string, 0)
+	for k := range getClusterNodeInfo() {
```
ondrej-fabry: Seems like this part of sorting by node name is duplicated in PrintAllIpams() as well, perhaps it should be extracted into some function like: `getSortedNodeInfo() []*node.NodeInfo`, that returns all nodes sorted by name.

ondrej-fabry: That's why I proposed new function `getSortedNodeInfo()` to be added next to the old `getClusterNodeInfo()` and use it only in those cases where it's needed, simply to avoid these two cases, in `PrintNodes()` and `PrintAllIpams()`, of repeated sorting code and have both of them shortened to:

```go
for _, n := range getSortedNodeInfo() {
    // ...
}
```

rastislavszabo: Merging to have this included in the release. Jan, please open another MR if you want to apply Ondrej's proposal.

PR #1193, opened 2018-10-19
Add the node config CRD writeup to the TOOLS.md page

jmedved:


docs/TOOLS.md
```diff
@@ -74,3 +74,54 @@ $ contiv-netctl vppdump k8s-master
...
+   kubectl apply -f <configuration-file>
+   ```
+   You can also get the current cluster configration as follows:
+   ```
+   kubectl get nodeconfigs.nodeconfig.contiv.vpp -o json
+   ```
+
+* `telemetryreports.telemetry.contiv.vpp`: provides telemetry data from
+  all Contiv-VPP vswitches in the cluster. The telemetry data is
+  basically a dump of VPP state and a eport on the health of the
```
lukasmacko: type `report` missing `r`

docs/TOOLS.md
```diff
@@ -74,3 +74,54 @@ $ contiv-netctl vppdump k8s-master
...
+## Contiv -VPP Custom Resource Definitions (CRDs)
+
+A resource is an endpoint in the [Kubernetes API][1] that stores a
+collection of [API objects][2] of a certain kind. For example, the
+built-in pods resource contains a collection of Pod objects. A [custom
+resource][3] is an extension of the Kubernetes API that is not necessarily
+available on every Kubernetes cluster. In other words, it represents a
+customization of a particular Kubernetes installation.
+
+Contiv-VPP defines the following custom resources:
```
brecode: Maybe we should add here that in order to use CRDs you need to: 
1. In a vagrant world edit the value of export CRD_DISABLED=${CRD_DISABLED:-true} to false found in this file 
https://github.com/contiv/vpp/blob/master/vagrant/defaults

2. In a manual setup use helm to generate the contiv-vpp.yaml with CRDs enabled
`$ helm template --name foo --set contiv.crdNodeConfigurationDisabled=false --set contiv.ipamConfig.contivCIDR=10.128.0.0/14 $cvpp_dir/k8s/contiv-vpp > $cvpp_dir/k8s/contiv-vpp/foo.yaml`

PR #1202, opened 2018-10-31
Fix an error of vppctl when it meets multiple contivvpp/vswitch images

TrevorTaoARM:
When there are multiple contivvpp/vswitch related images, such as
contivvpp/vswitch-amd64:latest, contivvpp/vswitch-amd64:xxx, and
contivvpp/vswitch:latest, the current vppctl utility may just
select the first image as the running image, as it would show the
error:
Error finding a running container derived from the contivvpp/vswitch image.

Now we do an exact match and select the correct one from the multiple
candidate contivvpp/vswitch image for which it fixes that error.


rastislavszabo:
Superseded by https://github.com/contiv/vpp/pull/1350

PR #1226, opened 2018-11-08
common memif mount point for vnfs/vswitch, 

jhnbrns:
The vswitch and vnfs need a common mount point such as /tmp or /var/run/contiv for the memif sockets, otherwise traffic will not pass.

Also avoid sfc/contiv colliding on gig-e config ... the sfc.conf requires a bypass_renderer: true to be set so that the controller does not write to the gige interfaces.

Also, care must be taken to ensure the config in the sfc.conf for the gige interfaces matches that of the system.  The sfc-controller will soon discover the contiv host gige config so these will not be required in the sfc.conf and thus wont have to be configured in two different places.


brecode:
LGTM, mergina

PR #1310, opened 2018-12-13
(dev) Update Dockers to ubuntu:18.04 & Go 1.10.5

rastislavszabo:


docker/vpp-vswitch/vpp/Dockerfile
```diff
@@ -1,18 +1,25 @@
...
-    make autoconf automake libtool curl unzip \
-    ethtool\
+    # general tools
+    sudo wget git ca-certificates \
+    # build tools
+    make \
+    # vpp-agent & vpp process manger
+    python supervisor \
+    # required for disabling TCP checksum offload in containers
+    ethtool \
```
milanlenco: Btw. checksum offloading is now disabled by vpp-agent via syscall, but I would keep ethtool installed anyway, just for debugging purposes.

PR #1341, opened 2019-01-10
(dev) Fix L2 mode, get rid of PodVPPSubnetCIDR

rastislavszabo:
validator needs to be updated, therefore merging into dev for now


plugins/ipv4net/ipv4net_api.go
```diff
@@ -28,6 +28,10 @@ type API interface {
 	// GetHostIPs returns all IP addresses of this node present in the host network namespace (Linux).
 	GetHostIPs() []net.IP
 
+	// GetPodSubnetThisNode returns POD network for the current node
+	// (given by nodeID allocated for this node).
+	GetPodSubnetThisNode() *net.IPNet
```
milanlenco: Why was this API added to ipv4net plugin? This information is already provided by IPAM and ipv4net is now just forwarding it. Any plugin that needs to know the pod subnet can add IPAM as dependency and read it directly from there.

plugins/service/renderer/nat44/nat44_renderer.go
```diff
@@ -502,22 +510,25 @@ func (rndr *Renderer) exportServiceIPMappings(service *renderer.ContivService,
...
 		}
 	}
-
 	return false
 }
 
+// isLocalPodIP returns true if the given IP is this node's local POD IP.
+func (rndr *Renderer) isLocalPodIP(ip net.IP) bool {
+	return rndr.IPv4Net.GetPodSubnetThisNode().Contains(ip)
+}
```
milanlenco: Renderer already has IPAM as dependency so you can do:
```
rndr.IPAM.GetPodSubnetThisNode().Contains(ip)
```

PR #1393, opened 2019-02-11
(dev) External IPAM (Calico-VPP)

rastislavszabo:


plugins/bgpreflector/bgpreflector.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Resync resynchronizes BGPReflector against the BGP routes in the Linux host.
+// A set of already allocated pod IPs is updated.
+func (br *BGPReflector) Resync(event controller.Event, kubeStateData controller.KubeStateData,
+	resyncCount int, txn controller.ResyncOperations) (err error) {
+
+	// do not initialize if external IPAM is not in use
+	if !br.ContivConf.GetIPAMConfig().UseExternalIPAM {
+		return nil
```
milanlenco: the condition can be removed - the case is excluded by HandlesEvent()

PR #1408, opened 2019-02-14
Manual installation updates

rastislavszabo:


jmedved:
This is a really good writeup. Let’s add contiv with this pointer to kubeadm (I think it’s a PR on the kubeadm man page)

PR #1418, opened 2019-02-19
IPv6 initial support

lukasmacko:


plugins/ipam/ipam.go
```diff
@@ -323,15 +325,23 @@ func (i *IPAM) NodeIPAddress(nodeID uint32) (net.IP, *net.IPNet, error) {
...
+		IP:   net.IP(ip).Mask(mask),
 		Mask: mask,
 	}
 	return nodeIP, nodeIPNetwork, nil
 }
 
+// IsNodeInterconnectIPv6 returns true if nodeInterconnectSubnet is IPv6 range
+func (i *IPAM) IsNodeInterconnectIPv6() bool {
+	return strings.Contains(i.ContivConf.GetIPAMConfig().NodeInterconnectCIDR.String(), ":")
+}
```
milanlenco: Don't forget also about ContivCIDR, which IPAM may use to calculate its own node interconnect CIDR (makes sense for IPv6?), i.e. perhaps use instead: 
`strings.Contains(i.nodeInterconnectSubnet.String(), ":")`

rastislavszabo: ContivCIDR does not work with IPv6 at this point, the IPAM plugin needs to be adapted

PR #1442, opened 2019-02-26
(dev) ipv6route service renderer skeleton

rastislavszabo:


plugins/service/renderer/ipv6route/ipv6route_renderer.go
```diff
@@ -0,0 +1,204 @@
...
+// AddService installs VPP config for a newly added service.
+func (rndr *Renderer) AddService(service *renderer.ContivService) error {
+	if rndr.snatOnly {
+		return nil
+	}
+
+	config := rndr.renderService(service)
+
+	txn := rndr.UpdateTxnFactory(fmt.Sprintf("add service '%v'", service.ID))
+	for key, val := range config {
```
milanlenco: just FYI: you can use PutAll and DeleteAll from controller/api to replace these for cycles (used for rendered config in AddPod, DeletePod)

rastislavszabo: thx :)

PR #1449, opened 2019-03-04
(dev)Fix repeated add of IP into node struct

lukasmacko:


plugins/ipv4net/resync_events.go
```diff
@@ -199,7 +199,11 @@ func (n *IPv4Net) configureMainVPPInterface(event controller.Event, txn controll
 	if len(n.nodeIP) > 0 {
 		nodeIPs = append(nodeIPs, &contivconf.IPWithNetwork{Address: n.nodeIP, Network: n.nodeIPNet})
 	}
-	n.NodeSync.PublishNodeIPs(nodeIPs, contivconf.IPv4)
+	ipVersion := contivconf.IPv4
+	if n.nodeIP.To4() == nil {
```
rastislavszabo: use already existing `isIPv6` function from `utils.go`

PR #1465, opened 2019-03-19
iptables policy renderer

lukasmacko:


plugins/policy/plugin_impl_policy.go
```diff
@@ -96,20 +102,38 @@ func (p *Plugin) Init() error {
...
+				},
+				ResyncTxnFactory: func() controller.ResyncOperations {
+					return p.resyncTxn
+				},
 			},
-			ResyncTxnFactory: func() controller.ResyncOperations {
-				return p.resyncTxn
+		}
+	} else {
+		p.iptablesRenderer = &iptables.Renderer{
```
milanlenco: Perhaps iptablesRenderer could be a configurable alternative for IPv4 as well (?) Might be interesting for performance comparisons, etc.

PR #1470, opened 2019-03-25
Egress

lukasmacko:


plugins/policy/renderer/iptables/iptables_renderer.go
```diff
@@ -49,9 +56,22 @@ type Deps struct {
...
+	cacheTxn cache.Txn
 	resync   bool
 }
 
+// Init initializes the ACL Renderer.
+func (r *Renderer) Init() error {
+	r.cache = &cache.RendererCache{}
+	if r.LogFactory != nil {
+		r.cache.Log = r.LogFactory.NewLogger("-aclCache")
+	} else {
```
milanlenco: update method comment + logger name to use "iptables" instead of "acl"

PR #1516, opened 2019-04-09
Expose contivconf & ipnet structs in separate API packages

rastislavszabo:
Signed-off-by: Rastislav Szabo <raszabo@cisco.com>

### DIFF
--- a/plugins/contivconf/contivconf_api.go
+++ b/plugins/contivconf/contivconf_api.go
@@ -17,6 +17,7 @@ import (
 	"net"
 
 	stn_grpc "github.com/contiv/vpp/cmd/contiv-stn/model/stn"
+	"github.com/contiv/vpp/plugins/contivconf/api"
 	controller "github.com/contiv/vpp/plugins/controller/api"
 )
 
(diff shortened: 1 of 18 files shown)


plugins/contivconf/contivconf_api.go
```diff
@@ -17,6 +17,7 @@ import (
 	"net"
```
milanlenco: Maybe contivconf_api.go could be moved to api sub-directory? (same with ipnet_api.go)

plugins/contivconf/contivconf_api.go
```diff
@@ -140,13 +141,13 @@ type CustomIPAMSubnets struct {
 	// This is subnet for all PODs across all nodes.
 	PodSubnetCIDR *net.IPNet
 
-	// Prefix length of subnet used for all PODs within 1 node.
+	// RESTPrefix length of subnet used for all PODs within 1 node.
```
milanlenco: Unintentional?

plugins/contivconf/contivconf_api.go
milanlenco: Going back to this - I find it quite confusing now that a subset of contivconf/ipnet APIs (_some_ data structures) are under api/ directory whereas interfaces (and _other_ data structures) remained in the plugin main directory. Dependant plugins may need to import both directories instead of just "api".

PR #1560, opened 2019-05-24
Support for SRv6 node-to-node communication in IPv4 pod environment

fgschwan:
Change has multiple parts:
1. configuration changes - needed a third option from bool noOverlay mode -> made nodeToNodeTransport that covers old options (vxlan, nooverlay) and new added option (srv6), also useSRv6Interconnect configuration renamed because node-to-node-communication part of it moved to nodeToNodeTransport  and what's left is related only to k8s services
(dafaults stayed the same, for more info see [k8s/README.md](https://github.com/contiv/vpp/compare/dev...fgschwan:dt4-cases?expand=1#diff-e42e703f8eacc6c0bc4465761e4cdc1b))
2. SRv6 node-to-node communication itself
3. refactoring 
4. covering managment IP traffic also under SRv6 node-to-node communication (missed that previously)


docs/dev-guide/SERVICES.md
```diff
@@ -519,17 +519,17 @@ metadata:
...
 ```
 `manifest.yaml` file is generated by Helm, so you can alternatively change the helm template 
 input values in [values.yaml][values.yaml] (or [values-arm64.yaml][values-arm64.yaml])
 ```
 ...
 contiv:
-  useNoOverlay: false
-  useSRv6Interconnect: false
+  nodeToNodeTransport: "nooverlay"
+  useSRv6ForServices: false
```
rastislavszabo: shouldn't this be true?

fgschwan: fixed and enhanced in https://github.com/contiv/vpp/pull/1560/commits/8d207a7c5cf9bb116955cef8f6a8be3c50b3494b

PR #1584, opened 2019-06-20
SFC Processor implementation + l2xconn SFC renderer skeleton

rastislavszabo:


rastislavszabo:
Hi @AdelBouridah 
yes exactly. Take the l2xconn renderer as a skeleton that needs to be implemented, not as an actual implementation :)

Issue #1164, opened 2018-10-09
Sporadic kubectl exec issue Ubuntu 18.04


rastislavszabo:
In the Contiv/VPP CI, we sporadically see the following issue by executing `kubectl exec ...`:

```
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:90: adding pid 12012 to cgroups caused \"failed to write 12012 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/kubepods/besteffort/podeb5707e6-cb5b-11e8-9c69-024d82c4d587/8f696a77036051847f87e9b0dd86d0ff44ec03073e5baebc199fe9d1fd8fe607/cgroup.procs: invalid argument\"": unknown
```

This started to appear on Ubuntu 18.04. No issues with Ubuntu 16.04.

rastislavszabo:
The root cause seems to be https://github.com/opencontainers/runc/issues/1884

rastislavszabo:
not reproducible anymore

rastislavszabo:
still present on k8s 1.14.2

PR #1611, opened 2019-07-18
Update vagrantfile to allow configuration of vpp worker threads

samelias:
Signed-off-by: samuel.elias <samelias@cisco.com>


rastislavszabo:
It seems that https://github.com/contiv/vpp/issues/1164 is still reproducible with this Ubuntu version. Can we try to incorporate a newer Docker version in the Vagrant image?

PR #1616, opened 2019-07-23
Add support for SRv6 END.DX4

kozemcak:
- Node1:Pod1 > SRv6 > Cluster:Service -> END.DX4

Signed-off-by: Andrej Kozemcak <andrej.kozemcak@pantheon.tech>

### DIFF
--- a/plugins/service/renderer/srv6/srv6_renderer.go
+++ b/plugins/service/renderer/srv6/srv6_renderer.go
@@ -45,6 +45,7 @@ const (
 )
 
 const (
+	ipv4HostPrefix   = "/32"
 	ipv6HostPrefix   = "/128"
 	ipv6PodSidPrefix = "/128"
 	ipv6AddrAny      = "::"
@@ -199,6 +200,38 @@ func (r *Renderer) Resync(resyncEv *renderer.ResyncEventData) error {
 	return nil
 }
 
+// isIPv6 returns true if the IP address is an IPv6 address, false otherwise.
+func isIPv6(ip net.IP) bool {
+	if ip == nil {
+		return false
+	}
+	return strings.Contains(ip.String(), ":")
+}
+
+func convertIPv4ToIPv6(ip net.IP) net.IP {
+	if ip == nil {
+		return nil
+	}
+
+	if isIPv6(ip) {
+		return ip
+	}
+
+	return ip.To16()
+}
+
+func getHostPrefix(ip net.IP) string {
+	if ip == nil {
+		return ""
+	}
+
+	if isIPv6(ip) {
+		return ipv6HostPrefix
+	}
+
+	return ipv4HostPrefix
+}
+
 // renderService renders Contiv service to VPP configuration.
 // addDelConfig sliceContains KV pairs that should be added/deleted,
 // updateConfig sliceContains KV pair that should be updated.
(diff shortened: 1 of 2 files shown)


k8s/contiv-vpp/values.yaml
```diff
@@ -3,6 +3,7 @@ contiv:
   nodeToNodeTransport: "vxlan"  # possible values: "vxlan", "srv6", "nooverlay"
   useSRv6ForServices: false
   useTAPInterfaces: true
+  srv6ServiceIP4: false
```
rastislavszabo: Do we really need this option? Wouldn't `useSRv6ForServices: true` be enough? In IPv4 setup, it would mean IPv4 services over SRv6, in IPv6 setup it would mean IPv6 services over SRv6

rastislavszabo: I don't actually see this being used in the code :)

kozemcak: Sorry, my fault. I forgot/miss remove this value.

rastislavszabo:
Just a small note, with this, the ACL-based network policy renderer probably won't work correctly. I am afraid that for SRv6 + IPv4, we would need to use the iptables renderer as we do for IPv6: https://github.com/contiv/vpp/blob/master/plugins/policy/plugin_impl_policy.go#L124-L139

plugins/service/renderer/srv6/srv6_renderer.go
```diff
@@ -339,14 +378,29 @@ func (r *Renderer) renderService(service *renderer.ContivService, oper operation
...
 		if !(oper == serviceDel && usedInOtherService) { // don't delete local sid if it is used in other service
-			localSID := &vpp_srv6.LocalSID{
-				Sid:               r.IPAM.SidForServicePodLocalsid(backend.ip).String(),
-				InstallationVrfId: r.ContivConf.GetRoutingConfig().PodVRFID,
-				EndFunction: &vpp_srv6.LocalSID_EndFunction_DX6{EndFunction_DX6: &vpp_srv6.LocalSID_EndDX6{
-					NextHop:           backend.ip.String(),
-					OutgoingInterface: vppIfName,
-				}},
+			var localSID *vpp_srv6.LocalSID
+			if isIPv6(r.ContivConf.GetIPAMConfig().PodSubnetCIDR.IP) {
```
fgschwan: This should probably be r.ContivConf.GetIPAMConfig().UseIPv6. It is basically the same, but it handles some special cases better:
1. not filled PodSubnetCIRD defaulting to IPv4
2. pod subnet CIDR can be also defined in one big string (ContivCIDR) together with other subnet CIDRs (i never used this, but it is possible)

And also IPv6/IPv4 code switches can be tracked easily by finding usage of this config variable.

plugins/service/renderer/srv6/srv6_renderer.go
```diff
@@ -199,6 +200,43 @@ func (r *Renderer) Resync(resyncEv *renderer.ResyncEventData) error {
...
+	}
+
+	if isIPv6(ip) {
+		return ip
+	}
+
+	ip6 := ip.To16()
+
+	ip6[11] = 0xff
+	ip6[10] = 0xff
```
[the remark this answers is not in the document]
fgschwan: fixed

plugins/service/renderer/srv6/srv6_renderer.go
fgschwan: fixed

PR #1626, opened 2019-07-30
Add test for srv6_renderer: IPv4 Pod, IPv6 Enviroment - END.DX4

kozemcak:
- test use case Pod/Service for IPv4 pod/IPv6 fabrict enviroment
  - Node1:Pod1 > SRv6 > Cluster:Service -> END.DX4

Signed-off-by: Andrej Kozemcak <andrej.kozemcak@pantheon.tech>

### DIFF
--- a/plugins/service/renderer/srv6/srv6_renderer_test.go
+++ b/plugins/service/renderer/srv6/srv6_renderer_test.go
@@ -446,12 +446,27 @@ func addServiceEndpoints(pod1 podmodel.ID, pod1IP net.IP, pod2 podmodel.ID, pod2
 }
 
 func addServiceEndpointsInMultipleNodes(pods []podEndPoints, port int32, data *data) {
+	address := make([]*epmodel.EndpointSubset_EndpointAddress, 0, len(pods))
+	for _, pod := range pods {
+		addr := &epmodel.EndpointSubset_EndpointAddress{
+			Ip:       pod.podIP.String(),
+			NodeName: pod.podNode,
+			TargetRef: &epmodel.ObjectReference{
+				Kind:      "Pod",
+				Namespace: pod.pod.Namespace,
+				Name:      pod.pod.Name,
+			},
+		}
+
+		address = append(address, addr)
+	}
+
 	eps1 := &epmodel.Endpoints{
 		Name:      "service1",
 		Namespace: renderer_testing.Namespace1,
 		EndpointSubsets: []*epmodel.EndpointSubset{
 			{
-				Addresses: make([]*epmodel.EndpointSubset_EndpointAddress, len(pods)),
+				Addresses: address,
 				Ports: []*epmodel.EndpointSubset_EndpointPort{
 					{
 						Name:     "http",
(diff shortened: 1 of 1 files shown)


plugins/service/renderer/srv6/srv6_renderer_test.go
```diff
@@ -79,7 +81,70 @@ type data struct {
 func TestBasicService(t *testing.T) {
 	RegisterTestingT(t)
 	retriever := configRetriever.NewMockConfigRetriever()
-	data := initTest("TestBasicService", defaultConfig(), retriever, false)
+	data := initTest("TestBasicService", defaultConfig(false), retriever, false)
+
+	// setup service
+	emptyResync(data)
+	service1 := addServiceMetadata(data, defaultPort)
+	pod1IP := addLocalPod(renderer_testing.Pod1, data, retriever)
```
fgschwan: This is confusing (apart from almost no change between this and ipv4 version of test that confused github compare functionality->commenting in ipv6 test instead of ipv4 test). The only difference is boolean parameter in defaultConfig. How is this related to IPv6 vs IPv4pods/IPv6 test? If you inspect initTest and then addLocalPod, it becomes clear. Changed config in data variable change pod IP address generation in addLocalPod and you have IPv4 IP address instead of IPv6 address. Could you make it more obvious by using pod1IPv4/pod2Ipv4 instead of pod1IP/pod2IP and add comment to this line (in ipv4 test) that IP generation depends on config that is IPv4-pod oriented for ipv4 test?

plugins/service/renderer/srv6/srv6_renderer_test.go
```diff
@@ -217,6 +282,32 @@ func TestServiceWithRemoteBackend(t *testing.T) {
...
 
+func TestServiceWithRemoteBackendIPv4(t *testing.T) {
+	RegisterTestingT(t)
+	retriever := configRetriever.NewMockConfigRetriever()
+	data := initTest("TestServiceWithRemoteBackend", defaultConfig(true), retriever, false)
+
+	// setup service
+	emptyResync(data)
+	service1 := addServiceMetadata(data, defaultPort)
+	pod2IP := addLocalPod(renderer_testing.Pod2, data, retriever)
```
fgschwan: In the light of my previous comment about pod IP creation, my first thought is to rename this too, but idea is to test remove backend and not local backend. So pod2 shouldn't be in test at all. This test is copy of TestServiceWithRemoteBackend where it is done this way (and i did that test that nonclean way). Now is probably that time to refactor and simplify both test by removing local backend. The main work will be in refactoring addServiceEndpointsInMultipleNodes so it supports also one endpoint and not fixed 2 endpoints.

plugins/service/renderer/srv6/srv6_renderer_test.go
kozemcak: Rename

plugins/service/renderer/srv6/srv6_renderer_test.go
kozemcak: Done

PR #1646, opened 2019-08-15
do not merge: add VPP patch fixing GSO-IPV6

samelias:
Signed-off-by: samuel.elias <samelias@cisco.com>


samelias:
patch was verified and merged into VPP master

PR #1663, opened 2019-09-06
SRv6 implementation of SFC renderer

fgschwan:


rastislavszabo:
Just a general ask for now, could you please make sure that line length of your code does not exceed 120 characters, so that the diffs can be reviewed on Github in more convenient way?

plugins/sfc/renderer/api.go
```diff
@@ -148,6 +149,11 @@ type PodSF struct {
 	// without further processing. Non-local pods contain logical names as they came from CRD.
 	InputInterface  string // name of the interface trough which the traffic enters the pod
 	OutputInterface string // name of the interface using which the traffic leaves the pod
+
```
rastislavszabo: cc @milanlenco 

I'm not sure about this API / naming: `InputInterface` vs  `InputInterfaceConfigName` (and the same for output) for the virtual interface name defined in CRD vs logical name used for config. We may think about something else here.

Maybe `InputInterface` & `OutputInterface ` could be a struct, with members `CrdName` & `LogicalName`, or something similar?

Anyway, I would not name the CRD as "configuration file" in the comment, that may be misleading. It may not necessarily be any file.

fgschwan: fixed in https://github.com/contiv/vpp/pull/1663/commits/291b06f116853535bb5c1f0a35b90e4fd9ed863b (sorry for big commit)

fgschwan:
> Just a general ask for now, could you please make sure that line length of your code does not exceed 120 characters, so that the diffs can be reviewed on Github in more convenient way?

I cut the lines so it is more convenient

vendor/github.com/ligato/cn-infra/logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -119,7 +119,7 @@ func (p *Plugin) Init() error {
 //   > curl -X PUT http://localhost:<port>/log/<logger-name>/<log-level>
 func (p *Plugin) AfterInit() error {
 	if p.HTTP != nil {
-		p.HTTP.RegisterHTTPHandler(fmt.Sprintf("/log/{%s}/{%s:debug|info|warn|error|fatal|panic}",
+		p.HTTP.RegisterHTTPHandler(fmt.Sprintf("/log/{%s}/{%s}",
```
rastislavszabo: have you upstreamed these changes to ligato/cn-infra ? Otherwise they will be overwritten by next dependency update.

plugins/ipam/ipam.go
```diff
@@ -1036,29 +1147,185 @@ func (i *IPAM) SidForNodeToNodeHostLocalsid(nodeIP net.IP) net.IP {
...
+	// compute SID as combination of configurable prefix and IP address of service function pod
+	return i.combineMultipleIPAddresses(
+		newIPWithPositionableMaskFromIPNet(prefix),
+		newIPWithPositionableMask(serviceFunctionPodIP, prefixMaskSize, 128-prefixMaskSize))
+}
+
+// UpdateExternalInterfaceIPInfo is notifying IPAM about external interfacew IP allocation
+func (i *IPAM) UpdateExternalInterfaceIPInfo(extif, vppInterface string, nodeID uint32, ipNet *net.IPNet,
+	isDelete bool) {
+	i.mutex.Lock()
```
rastislavszabo: is this mutex lock really needed? Is this func called outside of the main event loop?

vendor/github.com/ligato/cn-infra/logging/logmanager/plugin_impl_log_manager.go
fgschwan: No. These changes appear in the newer version of cn-infra. When i prepared commits for this PR, i took the newest version of cn-infra as target for resolving vendor differences, but i should have taken the version used in Contiv's dev branch to stay in sync with output of dep tool. Fixing it in https://github.com/contiv/vpp/pull/1663/commits/6f4bd149aa3146ab7cd42a2d36a55f6147253898.

fgschwan: PS: the change is [fix](https://github.com/ligato/cn-infra/commit/b24f34014e533a82a1c744e84b1a5367ff5ad9b6#diff-5ccefe7767d91cafe6fa6ecf8c385132) in cn-infra for unrelated stuff. It fixes possible logging problems for vpp-agent tools where you can set logging by setting string ("debug", "warn",...) , so it should not change working/notworking status of functionality in SRv6 (and i think in Contiv in general).

plugins/ipam/ipam.go
fgschwan: It is called only from main event loop (so from main thread), so it is probably not. But all API calls have this (probably not needed except for those used by [REST](https://github.com/contiv/vpp/blob/master/plugins/ipam/rest.go)). Removing of mutexes in API methods can be possible proposal for future refactoring, but i leave this change now as it is to preserve API method implementation uniformity.

plugins/ipam/ipam.go
```diff
@@ -1036,29 +1147,185 @@ func (i *IPAM) SidForNodeToNodeHostLocalsid(nodeIP net.IP) net.IP {
...
+	prefixMaskSize, _ := prefix.Mask.Size()
+
+	// compute SID as combination of configurable prefix and IP address of service function pod
+	return i.combineMultipleIPAddresses(
+		newIPWithPositionableMaskFromIPNet(prefix),
+		newIPWithPositionableMask(serviceFunctionPodIP, prefixMaskSize, 128-prefixMaskSize))
+}
+
+// UpdateExternalInterfaceIPInfo is notifying IPAM about external interfacew IP allocation
+func (i *IPAM) UpdateExternalInterfaceIPInfo(extif, vppInterface string, nodeID uint32, ipNet *net.IPNet,
```
milanlenco: Couldn't IPAM read (and resync) IP addresses of external interfaces like IPNet plugin does from `kubeStateData[extifmodel.Keyword]`?
This would help to get rid of the Update... method from the IPAM plugin API.
Ideally, the IPAM plugin should only have Allocate/Get methods in its API.

jgallo542: Done here [https://github.com/contiv/vpp/pull/1663/commits/f1dff08f7085a916ddaf28b47fab261b7a039432](url)

plugins/ipam/ipam.go
```diff
@@ -266,26 +287,31 @@ func (i *IPAM) Resync(event controller.Event, kubeStateData controller.KubeState
```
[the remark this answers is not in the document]
jgallo542: It's been done in a method where this happens for pod related maps. However it not being the best place I moved it right before resync is being done [https://github.com/contiv/vpp/pull/1663/commits/48953a4241171388df18cea9e91860ad66877961](url)

PR #1682, opened 2019-09-25
Minor SFC CRD API changes + CRD schema validation …

rastislavszabo:
CRD API changes:
 - `ExternalInterface` CRD: `vlan` and `ip` members are now lowercase (to match with the rest of CRD APIs which follow camelCase notation)
 - `CustomNetwork` CRD: `stub` constant for network type becomes `STUB`


plugins/crd/handler/externalinterface/external_interface_handler.go
```diff
@@ -114,3 +115,61 @@ func (h *Handler) nodeInterfaceToProto(nodeIf v1.NodeInterface) *model.ExternalI
 	}
 	return protoVal
 }
+
+// Validation generates OpenAPIV3 validator for custom network CRD
+func Validation() *apiextv1beta1.CustomResourceValidation {
```
milanlenco: change "validator for custom network CRD" to "validator for external interfaces"

plugins/crd/handler/customnetwork/custom_network_handler.go
```diff
@@ -98,10 +99,51 @@ func (h *Handler) customNetworkToProto(customNetwork *v1.CustomNetwork) *model.C
 		customNetworkProto.Type = model.CustomNetwork_L2
 	case "L3":
 		customNetworkProto.Type = model.CustomNetwork_L3
-	case "stub":
+	case "STUB":
```
milanlenco: Maybe it is github but I do not see the stub->STUB update reflected in the examples.
I guess it works anyway though, since CustomNetwork_STUB is used if the above switch has no match, correct?
Not sure if and where we could throw error when a pod's annotation is incorrect.

rastislavszabo: Yeah, this was related just to the CRD. Actually, I think that I will `STUB` custom network option completely, as there is no need for having multiple stub networks.

PR #1705, opened 2019-10-24
Bug correction: if an SFC or part of an SFC may be releazed in severe…

AdelBouridah:
…al nodes i.e. multiple  nodes with locals Pods that belong to the SFC --> Chain local SFs only for the node with the lower ID as always done for instance

Signed-off-by: Adel Bouridah <a.bouridah@centre-univ-mila.dz>

### DIFF
--- a/plugins/sfc/renderer/l2xconn/l2xconn_renderer.go
+++ b/plugins/sfc/renderer/l2xconn/l2xconn_renderer.go
@@ -138,8 +138,8 @@ func (rndr *Renderer) renderChain(sfc *renderer.ContivSFC, isDelete bool) (confi
 		}
 
 		if iface != "" && prevIface != "" {
-			if rndr.isNodeLocalSF(sf) && rndr.isNodeLocalSF(prevSF) {
-				// the two SFs (prevSF and SF) are local - Rgarding the node ID (lower or note) cross-connect the interfaces
+			if rndr.shouldChainLocalSFs(sf, prevSF) {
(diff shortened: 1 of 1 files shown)


plugins/sfc/renderer/l2xconn/l2xconn_renderer.go
```diff
@@ -139,10 +139,11 @@ func (rndr *Renderer) renderChain(sfc *renderer.ContivSFC, isDelete bool) (confi
 
 		if iface != "" && prevIface != "" {
 			if rndr.isNodeLocalSF(sf) && rndr.isNodeLocalSF(prevSF) {
-				// the two SFs (prevSF and SF) are local - cross-connect the interfaces
-				xconnect := rndr.crossConnectIfaces(prevIface, iface, sfc.Unidirectional)
-				rndr.mergeConfiguration(config, xconnect)
-
```
milanlenco: Can be simplified to one-level `if`:
```
if rndr.shouldChainLocalSFs(sf, prevSF) {
    xconnect := rndr.crossConnectIfaces(prevIface, iface, sfc.Unidirectional)
    rndr.mergeConfiguration(config, xconnect)
}  else if rndr.shouldChainToRemoteSF(prevSF, sf) {
...
```

Also comment `... - Rgarding the node ID (lower or note) cross-connect the interfaces` doesn't make much sense.

AdelBouridah: Yes its OK, I have committed again the right code

PR #1708, opened 2019-11-04
l2xrender SFC local SFs bug correction -previously Forgeted for dev b…

AdelBouridah:
…ranch and mismatched a little in master one- here clean code to be applicated on dev and master branch

Signed-off-by: Adel Bouridah <a.bouridah@centre-univ-mila.dz>


rastislavszabo:
This would just cause conflicts between branches. There is no need to commit fixes into both branches - they are periodically synced. Done in https://github.com/contiv/vpp/pull/1710

